PR #185, opened 2021-08-20
Temp fix for useDapp bug

cryptoseneca:
Temporary fix for [useDapp bug](https://github.com/EthWorks/useDApp/issues/289)

### DIFF
--- a/packages/nouns-webapp/src/hooks/useContractFunction__fix.ts
+++ b/packages/nouns-webapp/src/hooks/useContractFunction__fix.ts
@@ -1,0 +1,65 @@
+import { useCallback, useState } from 'react';
+import { usePromiseTransaction } from '@usedapp/core/dist/esm/src/hooks/usePromiseTransaction';
+import { Contract } from '@usedapp/core/node_modules/ethers';
+import { Web3Provider } from '@usedapp/core/node_modules/@ethersproject/providers';
+import { TransactionOptions, useEthers } from '@usedapp/core';
+import { Signer } from 'ethers';
+
+// Temporary fix: https://github.com/EthWorks/useDApp/issues/289
+
+export const useContractFunction__fix = (
+  contract: Contract,
+  functionName: string,
+  options?: TransactionOptions,
+) => {
+  const { library, chainId } = useEthers();
+  const [events, setEvents] = useState<Record<string, any> | undefined>(undefined);
+
+  const { promiseTransaction, state } = usePromiseTransaction(chainId, options);
+
+  const send = useCallback(
+    async (...args: any[]) => {
+      const contractWithSigner = connectContractToSigner(contract, options, library);
+      const sendPromise = contractWithSigner[functionName](...args).then(
+        (result: any): Promise<any> => {
+          // Need to add chainId here to prevent "TypeError: Unsupported Chain" error message
+          result.chainId = chainId;
+          return result;
+        },
+      );
+
+      const receipt: any = await promiseTransaction(sendPromise);
+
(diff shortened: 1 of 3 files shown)


packages/nouns-webapp/src/hooks/useContractFunction__fix.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      if (receipt) {
+        if (receipt.logs && receipt.logs.length > 0) {
+          setEvents(receipt.logs.map((log: any) => contract.interface.parseLog(log)));
+        } else {
+          setEvents([]);
+        }
+      }
+    },
+    [contract, functionName, chainId, promiseTransaction, library, options],
```
gitblamecarrot, line 41: The latest release has `library` in index `3` which may not be an issue but may be worth keeping in mind. 

https://github.com/EthWorks/useDApp/blob/%40usedapp/core%400.4.4/packages/core/src/hooks/useContractFunction.ts#L42

packages/nouns-webapp/src/wrappers/nounsAuction.ts
```diff
@@ -1,6 +1,6 @@
 import { useContractCall } from '@usedapp/core';
 import { BigNumber as EthersBN, utils } from 'ethers';
-import { Contract } from '@ethersproject/contracts';
+import { Contract } from '@usedapp/core/node_modules/ethers';
```
solimander, line 3: Hmm shouldn't need to do this. The types are compatible now because both are on version 5.4.x. May need to nuke your local node modules and run yarn if you're getting type errors.

packages/nouns-webapp/src/hooks/useContractFunction__fix.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (options?.signer) {
+    return contract.connect(options.signer as Signer);
+  }
+
+  if (library?.getSigner()) {
+    return contract.connect(library.getSigner());
+  }
+
+  throw new TypeError('No signer available in contract, options or library');
+};
```
solimander, line 65: nit: this can still be imported from `@usedapp/core`

packages/nouns-webapp/src/hooks/useContractFunction__fix.ts
```diff
@@ -0,0 +1,65 @@
+import { useCallback, useState } from 'react';
+import { usePromiseTransaction } from '@usedapp/core/dist/esm/src/hooks/usePromiseTransaction';
+import { Contract } from '@usedapp/core/node_modules/ethers';
```
solimander, line 3: Same as above. Shouldn't be required

packages/nouns-webapp/src/wrappers/nounsAuction.ts
cryptoseneca: You're right - nuked node modules and it worked 👍

packages/nouns-webapp/src/hooks/useContractFunction__fix.ts
```diff
@@ -0,0 +1,44 @@
+import { useCallback, useState } from 'react';
+import { usePromiseTransaction } from '@usedapp/core/dist/esm/src/hooks/usePromiseTransaction';
+import { connectContractToSigner } from '@usedapp/core';
+import { Contract } from '@ethersproject/contracts';
+import { TransactionOptions, useEthers } from '@usedapp/core';
```
solimander, line 5: nit: combine useDapp imports

solimander (approved):

PR #226, opened 2021-09-16
Update timer logic to check auction status less frequently

MrRyanFloyd:
Update timer logic to check auction status less frequently for react performance improvement. This was causing performance issues in party nouns.
Check auction status once every 30 seconds to prevent unnecessary react re-renders. Then check every second in the final minute of the auction.


cryptoseneca:
Hey @MrRyanFloyd. Thanks for suggesting the edit. One modification I'd request is setting the check to every second in the last 5 minutes instead of the last minute of the auction which is the timeframe where the auction can be extended.

PR #246, opened 2021-10-06
Update timer logic to check auction status less frequently

MrRyanFloyd:
Update timer logic to check auction status less frequently to limit component re-renders.

Initially check auction status once every 30 seconds. When less than 5 minutes remain in the auction, check the auction status every second.

** This PR addresses previous comments from @cryptoseneca and is a copy of closed PR #226


gitblamecarrot (approved):
Sounds like a good experiment to me. This'll just keep the AuctionActivity from repainting but the timer should still tick down.

Issue #275, opened 2021-11-01
Profile page showing incorrect state for Brave proposal and voting


lastpunk9999:
The latest proposal is being displayed as "Pending", when it should be "Active"
Nounders Nouns have voted for the proposal, but Noun 0 is displayed as "waiting"

![image](https://user-images.githubusercontent.com/83411264/139694887-cbbd452e-4e4e-4931-abd1-8e402d23b27a.png)

![image](https://user-images.githubusercontent.com/83411264/139694977-7cbba178-941e-4086-9d34-d96b3ddfc76c.png)
cc @bcjgit

bcjgit:
@lastpunk9999 I think the code I pushed in #279 should fix issue 2. I pinged solimander about issue 1 as it seems to be at the graph level. 

If you want, we can move ahead fixing issue 2 while issue 1 gets resolved. Up to you.

lastpunk9999:
fixed by #279

PR #279, opened 2021-11-03
Fix status bug

bcjgit:
# TLDR
Addresses issue discussed in:
https://github.com/nounsDAO/nouns-monorepo/issues/275

# Testing done

On my local, text seemed to now be updated to what I expected. It's hard to test the exact bug @lastpunk9999 mentioned because there are currently no proposals in that state.  That said, fairly sure this should fix it.


bcjgit:
@cryptoseneca would you have time to take a look at this? Its sort of stalled out in review

PR #291, opened 2021-11-14
Add test for buildSVG

thomasmhr:


volkyeth (approved):
took us a minute, but LGTM

PR #300, opened 2021-12-02
Add end time to auction and in time to proposals

gitblamecarrot:
This adds the ability to click on the auction timer to have it display the ending time for the local user.

endsat

It also adds an `in XX t` helper to the proposal times to help users understand when things happen.

votein


cryptoseneca (approved):
useful details 🔥

PR #301, opened 2021-12-03
Add `getNounSeedFromBlockHash` to `@nouns/assets`

lastpunk9999:
This request adds a `getNounSeedFromBlockHash` function which emulates the NounSeeder contract's on-chain implementation  for generating a pseudorandom Noun using an id and a block hash.


packages/nouns-assets/package.json
```diff
@@ -11,10 +11,15 @@
     "encode": "ts-node scripts/encode.ts",
     "clean": "rimraf dist tsconfig.build.tsbuildinfo",
     "build": "tsc -p tsconfig.build.json",
-    "prepare": "yarn build"
+    "prepare": "yarn build",
+    "test": "mocha -r ts-node/register \"test/**/*.test.ts\""
   },
   "devDependencies": {
+    "@ethersproject/bignumber": "^5.5.0",
+    "@ethersproject/solidity": "^5.5.0",
```
solimander, line 19: These will both need to be included in `dependencies` as they're consumed in the build output

packages/nouns-assets/src/utils.ts
```diff
@@ -32,3 +34,43 @@ export const getRandomNounSeed = (): NounSeed => {
...
+ * Emulate bitwise right shift and uint cast
+ * @param hex Hex representation of a number
+ * @param shiftAmount The amount to right shift
+ * @param uintSize The uint bit size to cast to
+ */
+export const hexShiftAndCast = (hex: string, shiftAmount: number, uintSize: number): string => {
+  hex = hex.replace(/0x/,'')
+  const start = hex.length - ((shiftAmount + uintSize)/4)
+  const end = hex.length - (shiftAmount/4)
+  return "0x" + hex.substring(start,end)
```
solimander, line 48: Nit: Can use ether's BigNumbers for this as well:

```ts
const shifted = BigNumber.from(hex).shr(shiftAmount).toHexString();
return `0x${shifted.substring(shifted.length - (uintSize / 4))}`;
```

solimander:
Looks good 👍 Small thing - Can you run `yarn format` and commit? That should be automated, but it looks like it isn't working correctly with the monorepo. Will take a look at that.

packages/nouns-assets/src/utils.ts
lastpunk9999: Un-nitted!

lastpunk9999:
new commits: use BigNumber for right shift, move bignumber and solidity to dependencies,

PR #306, opened 2021-12-07
Include Lido balance in Navbar treasury details and link to Etherscan token holdings page

### LINKED ISSUE #305: Include stETH in treasury balance
The treasury balance in the nav bar currently only displays ETH held, but as of this date ~10% of the treasury is held in stETH. 

Since stETH has a constant conversion to ETH, it might be good enough to convert stETH shares balance into ETH and combine that with the ETH balance to display a combined view of the treasury. The stETH shares balance includes validator fees so is an accurate representation of the ETH exposure of the DAO.

If additional assets are held by the treasury we can update our approach as needed.

lastpunk9999:
This adds NounsExecutor's Lido stETH balance to its ETH balance in the Navbar. It also links the treasury to Etherscan's token holdings page.

Resolves #305

### DIFF
--- a/packages/nouns-contracts/hardhat.config.ts
+++ b/packages/nouns-contracts/hardhat.config.ts
@@ -35,7 +35,7 @@ const config: HardhatUserConfig = {
         : [process.env.WALLET_PRIVATE_KEY!].filter(Boolean),
     },
     hardhat: {
-      initialBaseFeePerGas: 0, 
+      initialBaseFeePerGas: 0,
     },
   },
   etherscan: {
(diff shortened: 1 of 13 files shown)


solimander:
Couple small warnings turned CI error that are breaking the build:

```
Line 21:6:  React Hook useMemo has an unnecessary dependency: 'lidoToken'. Either exclude it or remove the dependency array. Outer scope values like 'lidoToken' aren't valid dependencies because mutating them doesn't re-render the component                  react-hooks/exhaustive-deps
Line 26:6:  React Hook useEffect has an unnecessary dependency: 'nounsDaoExecutor'. Either exclude it or remove the dependency array. Outer scope values like 'nounsDaoExecutor' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
```

solimander (approved):

PR #314, opened 2021-12-16
Proposal Creation Timestamp

solimander:
Add the proposal creation timestamp to the `Proposal` subgraph entity for use in the frontend.


bcjgit (approved):
🚢

Issue #357, opened 2022-02-10
Redirect /auction/:id => /noun/:id


bcjgit:
A user noted on [Discord](https://discord.com/channels/849745721544146955/913522194930364416/941396569599655936) that to maintain backwards compatibility we should ensure that `/auction/:id` links automatically redirect to `/noun/:id`. Currently, they go to the 404 page.

PR #378, opened 2022-02-22
Updates route to /auction backwards compatible with /noun

ryeshrimp:
Resolves https://github.com/nounsDAO/nouns-monorepo/issues/357


solimander (approved):
Thanks for the contribution!

PR #379, opened 2022-02-24
Updated proposal page

bcjgit:
This PR updates the UI of the proposal page (`/vote/:id`) to the new UI. 

# Testing done
1. Test all states via mocking boolean values
2. responsiveness checks (see note below)
3. [TODO] test on Rinkeby (finally got enough Rinkeby Nouns to make a test prop)

# Notes to reviewers

## Responsiveness 

There are 3 states to be aware of here:
1. xl (+1200px)
2. large (1199-992px)
3. Small (> 992px)

We had to introduce an extra state because in state 2 I opted to decrease the number of Nouns per row from 5 to 4. Without this, things looked strange. 

## Vote flow

As a part of this work, we had to redo the vote casting flow. It now works as follows (states detailed for ease of testing):

1. Submit vote button on Prop page only shows in enabled state if a user can vote (otherwise we show the appropriate null state copy when voting is active / don't show the button at all if voting is not active). 
2. Selecting the button leads to the following modal (zero votes shown here because of how I did the mocking)
Screen Shot 2022-02-24 at 3 36 31 PM
3. A user must select a vote for the submit vote button to be enabled
4. When a user selects a vote, the button retains its "active" state 
Screen Shot 2022-02-24 at 3 36 35 PM
5. Upon vote submit there are 2 cases: success or failure
6. Success modal (auto hides after 3 seconds):
Screen Shot 2022-02-24 at 3 38 26 PM
7. Failure (shows error message, does not a
Screen Shot 2022-02-24 at 3 36 07 PM
uto hide)

## Change state button

I also changed the style of the change state button
Screen Shot 2022-02-24 at 3 33 02 PM

## Past vote info

I also added a callout to display past voting per prop for connected wallet 

Screen Shot 2022-02-24 at 6 10 20 PM


packages/nouns-webapp/src/components/NounImageVoteTable/index.tsx
```diff
@@ -0,0 +1,42 @@
+import { StandaloneNounCircular } from '../../components/StandaloneNoun';
+import { BigNumber as EthersBN } from 'ethers';
+import classes from './NounImageVoteTable.module.css';
+import { GrayCircle } from '../GrayCircle';
+
+interface NounImageVoteTableProps {
+  nounIds: string[];
+}
+const NOUNS_PER_VOTE_CARD = 15;
```
cryptoseneca, line 9: nit: let's add `DESKTOP` to var name so we know that it's meant to be 15 on desktop only

cryptoseneca (commented):
Great job @bcjgit. Massive upgrade to the vote page here. Few things UX-wise:

## Noun card on mobile
The `NounImageVoteTable` feels like overkill on mobile. Maybe we can try reducing the space each voting card takes to have it each be 33% of one row?

mobile

## Noun card on desktop
- Spacing feels off
Screen Shot 2022-03-01 at 10 56 24 AM

## Error alert
- Copy feels too big, let's make it feel more cohesive with error message.
- Once error shows up, I can’t “reset state”. That is, after i dismiss the error alert and click “Submit vote” again, I get the same state. This happened when attempting to vote again (obvious error) but in production it might be a simple failed tx which would indeed require the user to submit again. 

Screen Shot 2022-03-01 at 10 46 26 AM

lastpunk9999:
1) Implement scrolling or another solution to see all the Nouns. It feels strange to see only the first 15 Nouns who voted when the count is usually much higher.
2) Use the proposal id as a seed to randomly shuffle the order of Nouns. I notice that Nounders Nouns most frequently show in the top 15 (and take up a lot of the slots) because we tend to vote at the beginning of the active period. If the Noun images were randomized it would allow more Nouns to be in the top 15.

lastpunk9999:
> 2. Use the proposal id as a seed to randomly shuffle the order of Nouns. 
Hacked some code to make this work: https://gist.github.com/lastpunk9999/5296a3ec8a41688e12b94f3c6d18cd62

bcjgit:
^^ Changes should address all feedback except for:

1. Scrolling Nouns (working on a solution here)
2. Making the vote counts 33% on mobile (open to doing it but went for the 1st of the suggested options where we just omit the Noun vote images on mobile)

cryptoseneca (commented):
Looks good! A few things:
- Shuffling of the nouns doesn't seem to be working.
- Can we go for 33% card for vote counts on mobile? While first option is better than full display, the six rows do take up too much real estate on mobile. Cutting this in half would make the whole feel more compact. 

### Alerts
While the decrease in fort size helps with the error alert, it still feels off. It might be due to the fact that the content ("well this is awkward") uses the same font as the title ("proposal #41") and is more or less still the same size. More broadly, I think it's more of a styling issue with the bootstrap alert component. What do you think about using our own `Modal` for both voting and alerts? It already comes with spacing/fonts/styling and plays well with mobile.

eg

lastpunk9999:
> * Shuffling of the nouns doesn't seem to be working

I think it works on my side @cryptoseneca 

On `Prop 39` without shuffling all Nounder Nouns are sequentially clustered together (Gnome/Noun 0, Shoe/Noun 10, etc)
image

With shuffling they are distributed
image

lastpunk9999 (changes requested):
@bcjgit I'm noticing a bug: when clicking on a Noun image in the proposal page, the app does not navigate to the correct Noun, but the last Noun that was viewed (eg: if I start on the current auction page, then navigate to a proposal, clicking a Noun brings me back to the current auction page)

bcjgit:
> @bcjgit I'm noticing a bug: when clicking on a Noun image in the proposal page, the app does not navigate to the correct Noun, but the last Noun that was viewed (eg: if I start on the current auction page, then navigate to a proposal, clicking a Noun brings me back to the current auction page)

Good catch. I think I know why that is. Fixing now...

lastpunk9999:
> 1. Scrolling Nouns (working on a solution here)

@bcjgit I wonder if scrolling within an element is awkward and maybe a modal would be better or easier?

bcjgit:
@lastpunk9999 

I fixed the issue you mentioned where noun images were not linking correctly. Per our discussion on Discord, I'll punt Noun scrolling / modal etc. to a follow up task (issue: https://github.com/nounsDAO/nouns-monorepo/issues/385)

@cryptoseneca 
1. Mobile vote count cards are not 33% per your comment
2. I migrated the vote modal to use the Nouns custom modal vs bootstrap modal
3. I've decrease the text size of the success / failure states in the vote modal to match with the modal style vs being large font + brand text 
4. As both 9999 and I do see Nouns shuffling  I took no action on that point ... let me know if it still isn't working for you

lastpunk9999 (approved):

cryptoseneca:
> @lastpunk9999
> 
> I fixed the issue you mentioned where noun images were not linking correctly. Per our discussion on Discord, I'll punt Noun scrolling / modal etc. to a follow up task (issue: #385)
> 
> @cryptoseneca
> 
> 1. Mobile vote count cards are not 33% per your comment
> 2. I migrated the vote modal to use the Nouns custom modal vs bootstrap modal
> 3. I've decrease the text size of the success / failure states in the vote modal to match with the modal style vs being large font + brand text
> 4. As both 9999 and I do see Nouns shuffling  I took no action on that point ... let me know if it still isn't working for you

my mistake, i thought shuffling was supposed to happen on each reload. i now see it's on a per proposal basis.

> @lastpunk9999
> 
> I fixed the issue you mentioned where noun images were not linking correctly. Per our discussion on Discord, I'll punt Noun scrolling / modal etc. to a follow up task (issue: #385)
> 
> @cryptoseneca
> 
> 1. Mobile vote count cards are not 33% per your comment
> 2. I migrated the vote modal to use the Nouns custom modal vs bootstrap modal
> 3. I've decrease the text size of the success / failure states in the vote modal to match with the modal style vs being large font + brand text
> 4. As both 9999 and I do see Nouns shuffling  I took no action on that point ... let me know if it still isn't working for you

my mistake. i thought shuffling was supposed to happen on each reload of page. i now see it's per proposal.

cryptoseneca (approved):

cryptoseneca (approved):

Issue #365, opened 2022-02-13
Use arrow keys to page though Nouns


bcjgit:
This has been an often requested feature. One example from discord:

https://discord.com/channels/849745721544146955/877215225093451776/942417178387419136

PR #380, opened 2022-02-25
Use arrow keys to page through Nouns page

ryeshrimp:
Resolves https://github.com/nounsDAO/nouns-monorepo/issues/365


lastpunk9999 (changes requested):
Thanks for this. The arrow navigation code works but I found a bug in the prev/next handlers that the arrow navigation exposes. If you are on the current Noun auction page and press the right arrow key several times, then hit the left key, the page does not navigate to the previous auction. This is because either the `handleKeyPress` function or the prev/next handlers don't check if they are allowed to navigate based on the current Noun ID. 
It might be a good idea to put some logic in one of these handlers to stop navigation in a similar way that the navigation UI buttons stop navigation.

ryeshrimp:
@lastpunk9999 Nice catch, I knew something was feeling weird when testing but I couldn't pinpoint what it was until you mentioned this. I chose to add it in the `handleKeyPress` function instead of the `prevAuctionHandler/nextAuctionHandler` to not duplicate code. Plus, these handlers were protected in other components against the same logic check.

packages/nouns-webapp/src/components/Auction/index.tsx
```diff
@@ -43,16 +43,22 @@ const Auction: React.FC<AuctionProps> = props => {
     currentAuction && history.push(`/noun/${currentAuction.nounId.toNumber() + 1}`);
   };
 
-  const handleKeyPress = useCallback(event => {
-    if (event.code === 'ArrowLeft') prevAuctionHandler();
-    else if (event.code === 'ArrowRight') nextAuctionHandler();
-    else return;
-  }, []);
+  const handleKeyPress = (event: KeyboardEvent) => {
```
lastpunk9999, line 46: The current code will cause adding and removing the event listener on every render because this function is being re-created on each render. Take note of the compiler warning about this:
```
src/components/Auction/index.tsx
  Line 46:9:  The 'handleKeyPress' function makes the dependencies of useEffect Hook (at line 61) change on every render. Move it inside the useEffect callback. Alternatively, wrap the definition of 'handleKeyPress' in its own useCallback() Hook
```

packages/nouns-webapp/src/components/Auction/index.tsx
```diff
@@ -43,16 +43,22 @@ const Auction: React.FC<AuctionProps> = props => {
...
+    if (currentAuction && lastNounId) {
+      const isFirstAuction = currentAuction.nounId.eq(0);
```
lastpunk9999, line 48: `isFirstAuction` and `isLastAuction` are used in a few other places, part of the solution might be to memoize these and reuse the output

packages/nouns-webapp/src/components/Auction/index.tsx
```diff
@@ -43,16 +43,22 @@ const Auction: React.FC<AuctionProps> = props => {
...
+      const isLastAuction = currentAuction.nounId.eq(lastNounId);
+
+      if (event.code === 'ArrowLeft' && !isFirstAuction) prevAuctionHandler();
```
lastpunk9999, line 51: It's also very likely you'll have to wrap `[prev/next]AuctionHandler` in a `useCallback` if they become dependencies of the key press effect.

lastpunk9999 (changes requested):
Good step toward a solution! This approach introduced a few other issues to tackle. Thanks for the submission!

packages/nouns-webapp/src/components/Auction/index.tsx
ryeshrimp: I removed useCallback because it was causing some issues but I'll take another look.

PR #387, opened 2022-03-08
Dynamically set the homepage document title based on the time remaining

lastpunk9999:
Sets the document title to the auction countdown timer if it's toggled on. If it's off, set the title to the end time of the auction. The countdown timer displays the seconds remaining only when the auction has less than 5 minutes remaining.

**Greater than 5 minutes remaining and timer toggled on**
image

**Less than 5 minutes remaining and timer toggled on**
image

**Timer toggled off**
image

### DIFF
--- a/packages/nouns-webapp/src/components/AuctionTimer/index.tsx
+++ b/packages/nouns-webapp/src/components/AuctionTimer/index.tsx
@@ -5,6 +5,7 @@ import classes from './AuctionTimer.module.css';
 import { useState, useEffect, useRef } from 'react';
 import { Row, Col } from 'react-bootstrap';
 import { useAppSelector } from '../../hooks';
+import useDocumentTitle from '../../hooks/useDocumentTitle';
 import clsx from 'clsx';
 
 dayjs.extend(duration);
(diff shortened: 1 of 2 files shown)


packages/nouns-webapp/src/components/AuctionTimer/index.tsx
```diff
@@ -45,11 +46,22 @@ const AuctionTimer: React.FC<{
...
   const auctionContentLong = auctionEnded ? 'Auction ended' : 'Auction ends in';
   const auctionContentShort = auctionEnded ? 'Auction ended' : 'Time left';
-
+  const flooredHours = Math.floor(timerDuration.hours());
   const flooredMinutes = Math.floor(timerDuration.minutes());
   const flooredSeconds = Math.floor(timerDuration.seconds());
   const isCool = useAppSelector(state => state.application.isCoolBackground);
 
+  // Dynamically sets the document title based on `timerToggler` and `auctionTime` remaining
+  let docTitle = `${flooredHours}h ${flooredMinutes}m`;
```
cryptoseneca, line 55: nit: move logic outside of component

```
const docTitle = (
  hours: number,
  minutes: number,
  seconds: number,
  timer: number,
  endTime: Dayjs,
): string => {
  let docTitle = `${hours}h ${minutes}m`;
  if (timer) {
    // Update every second when time remaining is less than 5 minutes
    if (timer < 300) docTitle = `${docTitle} ${seconds}s`;
  } else {
    docTitle = endTime.format('h:mm:ss a');
  }
  return (docTitle = `(${docTitle})`);
};
```

bcjgit (approved):
@lastpunk9999 I love this idea!

I am willing to ship "as is" but wanted to throw out one idea as well:

It feels potentially a bit confusing to have the time in the title with no context. I wonder if we could change the copy of the title (on the auction page) to reflect this more clearly. Something like Nouns - auction in <time>. Not sure if that would even fit honestly.  Just a thought.

cryptoseneca (commented):
There is an issue when tab becomes inactive where, due to low priority by browser, the timer goes out of sync. I ran into the same issue when building the initial auction timer. I was able to work around it due to the timer itself not being needed when a user is inactive. iirc, the way to fix is using web workers.

https://user-images.githubusercontent.com/85328329/157509908-e6d45d37-d1bc-45f9-bbdd-0ebb86dbc5f7.mov

lastpunk9999:
> There is an issue when tab becomes inactive where, due to low priority by browser, the timer goes out of sync. I ran into the same issue when building the initial auction timer. I was able to work around it due to the timer itself not being needed when a user is inactive. iirc, the way to fix is using web workers.

Well, that kind of goes against the whole reason for the PR which was to keep an eye on auction end times while working on other things. Maybe this code should be shelved for now unless I can come up with the active tab work around.

Issue #392, opened 2022-03-19
When not eligible to vote, still able to vote


bpeters:
Not only that but the TX will fail but the front-end will log the vote for the user.
<img width="971" alt="Screen Shot 2022-03-19 at 3 27 50 PM" src="https://user-images.githubusercontent.com/618157/159137476-914c4cf7-9607-4ad6-b400-c60ff47e1dc8.png">

PR #416, opened 2022-04-04
Vote Page Bug Fixes

### LINKED ISSUE #391: Prop 30 displays incorrect "against" vote count
https://nouns.wtf/vote/30 displays 2 votes against the proposal, when it should be 5. 2 is the number of owners of the 5 Nouns at time of voting.

This bug was present before the new UI introduced in PR #379 as seen in this screenshot
image

solimander:
This PR resolves #391 and #392.

For #391, the vote counts were correct, but the Noun images were not. We were pulling current Noun delegate information, rather than details from the time the proposal was created. The quick fix for this issue requires two calls - one to fetch the voter ids and a second time-travel request to fetch the Noun information from the proposal creation block. This can be reduced to a single call via a larger change to the subgraph in the future.

### DIFF
--- a/packages/nouns-webapp/src/utils/getNounsVotes.ts
+++ b/packages/nouns-webapp/src/utils/getNounsVotes.ts
@@ -1,4 +1,4 @@
-interface NounVote {
+interface Vote {
   supportDetailed: 0 | 1 | 2;
   nounsRepresented: string[];
 }
@@ -10,6 +10,6 @@ interface NounVote {
  * @param supportDetailed - The integer support value: against (0), for (1), or abstain (2)
  * @returns - flat list of nounIds that voted supportDetailed for the given prop
  */
-export const getNounVotes = (votes: NounVote[], supportDetailed: number) => {
+export const getNounVotes = (votes: Vote[], supportDetailed: number) => {
   return votes.filter(v => v.supportDetailed === supportDetailed).flatMap(v => v.nounsRepresented);
 };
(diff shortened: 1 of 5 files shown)


packages/nouns-webapp/src/wrappers/subgraph.ts
```diff
@@ -19,6 +19,28 @@ export interface IBid {
...
+  voter: {
+    id: string;
+  };
+}
+
+export interface ProposalVotes {
+  votes: ProposalVote[];
+}
+
+export interface DelegateNouns {
```
bcjgit, line 33: Non-blocking nit:

Assuming I'm understanding this correctly this represents 1 address (`id`) to whom a series of nouns has been delegates (`nounsRepresented`). If that's true, I wonder if the name could be changed to make it clear this is the delegated address vs being a Noun or something.

packages/nouns-webapp/src/utils/getNounsVotes.ts
```diff
@@ -1,14 +1,15 @@
+interface NounVote {
```
bcjgit, line 1: nit:

Similar comment to the one I made on the `DelegateNouns` interface ... this isn't _really_ a Noun vote right? It's the votes cast by a delegate with the represented Nouns ... right? Again, not blocking but wonder if naming could make this clearer.

bcjgit (approved):
Left a few non-blocking comments but overall lgtm. Thanks for this!

Issue #417, opened 2022-04-06
Make post auction win "get verified" flow easier to understand


bcjgit:
Many auction winners have expressed confusion on how to get verified after they win the auction. 

We should replace the get verified page with instructions on how to get verified in discord. 

discord context: https://discord.com/channels/849745721544146955/855538581434007635/961179363523588106

solimander:
Closed in #419

PR #419, opened 2022-04-09
Update "Get verified" button + Remove verify page

bcjgit:
Relevant issue: #417 

# TLDR
This PR does 2 things
- Replaces  the "Get verified" button with a "What now?" button that links to https://nouns.center/nouners
- Deprecates the `Verify` page (remove page, remove from routing, remove relevant components)

# Screenshots 

## Mobile
Screen Shot 2022-04-08 at 5 07 40 PM

## Desktop 
Screen Shot 2022-04-08 at 5 07 23 PM

# Testing done

- Hardcoded auction ended and winner variables to force winner verify treatment to show
- confirmed button looked right, confirmed that it opened relevant nouns.center page in a new tab 

# Note to reviewers

You'll notice I also made a small change on the Playground page. This was to fix an issue that seems to be the result of upgrading `@nouns/sdk` to the latest version. Happy to move this to its own PR if we think that'd have value.


packages/nouns-webapp/src/pages/Playground/index.tsx
```diff
@@ -56,7 +56,7 @@ const nounsSDKLink = (
 
 const DEFAULT_TRAIT_TYPE = 'heads';
 
-const encoder = new PNGCollectionEncoder(ImageData.palette);
+const encoder = new PNGCollectionEncoder();
```
solimander, line 59: `ImageData.palette` must be passed into the encoder. Otherwise, playground Nouns will be all black.

bcjgit: Whoops. I was def not using my brain when I made this change. My bad. Rectified.

solimander (approved):

packages/nouns-webapp/src/pages/Playground/index.tsx
solimander: No worries. If you made this change because you were getting a type error, rebuilt from the monorepo root to fix it.

PR #429, opened 2022-04-19
Add basic content moderation for ens names

bcjgit:
This pr adds logic to fallback to showing just ETH address when and ens matches a suite of regexes of unsavory words / phrases. 

## Before 
Screen Shot 2022-04-19 at 9 31 27 AM

## After

Screen Shot 2022-04-19 at 9 37 17 AM

# Some longer term thoughts

This is just a low hanging fruit solution. There are many ways a motivated attacker could still troll us. In particular, one attack vector here is that by having this logic in a public github repo we are showing all our cards so to speak. Probably not worth it in the near term but if moderation becomes a problem it might make sense to push this logic to a private api.

### DIFF
--- a/packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
+++ b/packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
@@ -1,4 +1,5 @@
 import moderationRegexes from './moderationRegexes.json';
+import Filter from 'bad-words';
 
 /**
  * Check if text matches a blocked phrase in a given langugae
@@ -12,14 +13,20 @@ export const containsBlockedText = (text: string, language: string) => {
   const regexesForLanguage = new Map(Object.entries(moderationRegexes)).get(language);
   // Default to letting the string through if the language is unsupprted
   if (regexesForLanguage === undefined) {
-    console.log(`Unsupported langugae ${language} requested`);
+    console.log(`Unsupported language ${language} requested`);
     return false;
   }
 
-  // Filter if we match at least one regex
+  // Filter based on bad-words profanity filters
+  const filter = new Filter();
+  if (filter.isProfane(text)) {
+    return true;
+  }
+
+  // Filter based on custom regexes
   return (
     regexesForLanguage
-      .map((entry: any) => {
+      .map((entry: { regex: string }) => {
         const regex = entry.regex;
         return text.match(regex) !== null;
       })
--- a/packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
+++ b/packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
@@ -1,25 +1,28 @@
-import moderationRegexes from "./moderationRegexes.json";  
+import moderationRegexes from './moderationRegexes.json';
 
 /**
  * Check if text matches a blocked phrase in a given langugae
- *  
- * @param text 
+ *
+ * @param text
  * @param language
  * @returns boolean true iif text matches a blocked word of phrase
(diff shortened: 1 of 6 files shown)


packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
```diff
@@ -0,0 +1,28 @@
+import moderationRegexes from './moderationRegexes.json';
+
+/**
+ * Check if text matches a blocked phrase in a given langugae
+ *
+ * @param text
+ * @param language
+ * @returns boolean true iif text matches a blocked word of phrase
+ */
+export const containsBlockedText = (text: string, language: string) => {
```
solimander, line 10: Thoughts on being slightly more proactive and using https://github.com/web-mech/badwords as a base with some custom words included. The fallback of showing an address makes accidental detection very low-stakes.

packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
```diff
@@ -0,0 +1,28 @@
...
+  // Get modearation regexes for language
+  const regexesForLanguage = new Map(Object.entries(moderationRegexes)).get(language);
+  // Default to letting the string through if the language is unsupprted
+  if (regexesForLanguage === undefined) {
+    console.log(`Unsupported langugae ${language} requested`);
```
solimander, line 15: `langugae` -> `language`

packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
```diff
@@ -0,0 +1,28 @@
...
+    return false;
+  }
+
+  // Filter if we match at least one regex
+  return (
+    regexesForLanguage
+      .map((entry: any) => {
```
solimander, line 22: Do not type as `any`

packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (regexesForLanguage === undefined) {
+    console.log(`Unsupported language ${language} requested`);
+    return false;
+  }
+
+  // Filter based on bad-words profanity filters
+  const filter = new Filter();
+  if (filter.isProfane(text)) {
+    return true;
+  }
```
solimander, line 24: may want to move this before the `regexesForLanguage` check. still want to check against the list even if no custom regexes exist.

packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
```diff
@@ -0,0 +1,35 @@
+import moderationRegexes from './moderationRegexes.json';
+import Filter from 'bad-words';
+
+/**
+ * Check if text matches a blocked phrase in a given langugae
+ *
+ * @param text
+ * @param language
+ * @returns boolean true iif text matches a blocked word of phrase
```
solimander, line 9: nit: `iif` -> `if`

packages/nouns-webapp/src/utils/moderation/containsBlockedText.ts
```diff
@@ -0,0 +1,35 @@
+import moderationRegexes from './moderationRegexes.json';
+import Filter from 'bad-words';
+
+/**
+ * Check if text matches a blocked phrase in a given langugae
```
solimander, line 5: nit: `langugae` -> `language`

solimander (approved):

PR #430, opened 2022-04-20
[DO NOT MERGE YET] Add Lingui.js to nouns.wtf + localize all strings

bcjgit:
# TLDR
This pr does three things:
1. Adds `lingui.js` to nouns.wtf (enabled for `en`, `ja`, and `pseduo-LOCALE` locales)
2. Wraps (nearly) all strings in `<Trans>` macro for translation
3. Includes the `*.po` files from the `lingui.js` compilation process to be passed onto translators (probably via Crowdin)

# Testing Done
I used pseudo-localization and visually confirmed (nearly) all strings were i18n-ed. 

# Notes to reviews

As I have been eluding to, there are a few strings that I did wrap.  

1. Anything in a dropdown menu (Playground has a lot of these, there is one in the create proposal flow)
 - The reason is that the Trans macro doesn't work with this because of how React works. There are some work arounds but I think they don't work in typescript. I am still looking into this
 2. A handful of what I deemed to be proper nouns (e.g. names of Github packages)
 3. Some Bootstrap form components that also didn't support non-string parameters (in particular,  file upload components used in the Playground and create proposal flows)

Other than these cases I *think* I got every string (all 221 of them by my count).


bcjgit:
Replaced by #433

PR #434, opened 2022-04-27
chore: update sharp to 0.30.x (fixes Mac M1 support)

marwanhilmi:
This fixes compatibility with Mac M1 and node v17.x. The current version (0.28.3) does not install and precompiled M1 / ARM binaries were only included in > 0.29.x.

- https://github.com/lovell/sharp/releases
- https://sharp.pixelplumbing.com/changelog


volkyeth:
those packages were discontinued on #891

PR #436, opened 2022-05-04
Rinkeby deploy of name reverted contracts + subgraph

thebuidler:
# Overview
This PR was primarily to revert the collapse from Noun(s) --> B01. Plural and singular were collapsed and the more refactoring and building we did on top of Nouns, the harder it would have been to change our name again.

I modified contracts and subgraph (which are the two packages that we'd more formally updated) + all other packages to use Noun(s) again until we formally decide on name. There might be some oddities in the two packages we added B01 support to **but** all tests pass.

As a means of testing + to better meet parity with Nouns, I've deployed all contracts to Rinkeby as well as the subgraph.

# Testing
* All contract tests pass
* Contracts deployed
* AH unpaused and mint + transferred correctly
* AH bid works as expected
* Subgraph properly creating entities


thebuidler:
Wow this was not meant for NounsDAO, my bad! 🤣

PR #438, opened 2022-05-06
[NIP-1] Dynamic quorum and proposal creation block fix

eladmallel:
## Overview

This PR is a continuation of the POC @solimander did in another branch. It introduces the new DAO logic contract `NounsDAOLogicV2`. High level changes in V2:

### Dynamic quorum

Proposals created in V2 will require more for votes as a function of their against votes. For full detail see the [spec](https://github.com/nounsDAO/nouns-tech/blob/main/dynamic-quorum/README.md), which includes a [demo spreadsheet](https://docs.google.com/spreadsheets/d/1trmtpgSLx8cfm-wD79mQWyh2y_GwW3X7f2rEnSWJ3AM/edit#gid=999443709). Here we'll just note the quorum function has useful variables the DAO can configure to adjust it at any time:

- Lower and upper bounds (in basis points)
- Polynom coefficients acting as nobs on how fast the quorum votes value grows
- An offset input that can delay when the function's boost kicks in (in basis points)

### Creation block fix

The DAO uses a votes snapshot at a proposal's *creation* block; however, in V1 we did not store that block, but rather upon vote casting, creation block was calculated thus: `proposal.startBlock - votingDelay`. This created the problem that if you change `votingDelay`, you change the block used to snapshot votes.

The solution in V2 is to store `creationBlock` in each proposal, so `votingDelay` is now only used during proposal creation, making it safe to change its value without risking in-progress proposals.


packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,835 @@
...
+    function _setMinQuorumVotesBPS(uint16 newMinQuorumVotesBPS) external {
+        DynamicQuorumParams memory params = getDynamicQuorumParamsAt(block.number);
+        params.minQuorumVotesBPS = newMinQuorumVotesBPS;
+        _setDynamicQuorumParams(params);
+    }
+
+    function _setMaxQuorumVotesBPS(uint16 newMaxQuorumVotesBPS) external {
+        DynamicQuorumParams memory params = getDynamicQuorumParamsAt(block.number);
+        params.maxQuorumVotesBPS = newMaxQuorumVotesBPS;
+        _setDynamicQuorumParams(params);
```
solimander, line 593: Feels like it may be better to split out validation and events to avoid revalidating and emitting unchanged params. Thoughts?

i.e.

```solidity
function _setMaxQuorumVotesBPS(uint16 newMaxQuorumVotesBPS) external {
    require(msg.sender == admin, 'NounsDAO::_setMaxQuorumVotesBPS: admin only');
    require(
        newMaxQuorumVotesBPS <= MAX_QUORUM_VOTES_BPS_UPPER_BOUND,
        'NounsDAO::_setMaxQuorumVotesBPS: invalid max quorum votes bps'
    );
    DynamicQuorumParams memory params = getDynamicQuorumParamsAt(block.number);

    uint16 oldMaxQuorumVotesBPS = params.maxQuorumVotesBPS;
    params.maxQuorumVotesBPS = newMaxQuorumVotesBPS;

    _writeQuorumParamsCheckpoint(params);

    emit MaxQuorumVotesBPSSet(oldMaxQuorumVotesBPS, newMaxQuorumVotesBPS);
}
```

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,835 @@
...
+            newProposal.id,
+            msg.sender,
+            targets,
+            values,
+            signatures,
+            calldatas,
+            newProposal.startBlock,
+            newProposal.endBlock,
+            newProposal.proposalThreshold,
+            newProposal.minQuorumVotes,
```
solimander, line 250: Thinking more about this, seems beneficial to provide the quorum votes value at the time of proposal creation (`minQuorumVotes()`), rather than `0`.

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
```diff
@@ -214,6 +227,176 @@ contract NounsDAOStorageV1 is NounsDAOProxyStorage {
...
+    struct DynamicQuorumParams {
+        /// @notice Minimum quorum votes BPS at the time of proposal creation
+        uint16 minQuorumVotesBPS;
+        /// @notice The maximum basis point number of votes in support of a proposal required in order for a quorum to be reached and for a vote to succeed. *DIFFERS from GovernerBravo
+        uint16 maxQuorumVotesBPS;
+        /// @notice The quorum votes polynom input offset which suppresses polynom contribution until againstVotes.div(totalSupply) reaches this value
+        uint16 quorumVotesBPSOffset;
+        /// @notice Polynomial coefficients for calculating a dynamic quorum based on the amount of against votes
+        /// @dev The coefficients are assumed to be fixed point integer with 6 decimals, i.e 0.2 is represented as 0.2 * 1e6 = 200000. There are 2 coefficients: x^1 and x^2
+        uint32[2] quorumPolynomCoefs;
```
solimander, line 355: We should consider converting the coefficient array to two `uint32`s. Don't see much benefit in the array over two separate variables, and this would allow the struct to pack into one slot, rather than two.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
solimander: Not thinking of gas savings here. Just seems a bit cleaner to reflect the exact state changes in the events, rather than emitting the same event multiple times in the same block if updating 2+ values.

eladmallel: @solimander makes sense, so we will add a task to refactor this code + update the subgraph code to index these individual events as well.

solimander: 🙏

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
eladmallel: Pushed a fix

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: Pushed a fix on the smart contract side; subgraph update still TODO

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
davidbrai: @solimander are you sure it doesn't pack it into one slot?
my understanding is that only dynamic arrays start a new storage slot (from reading [this](https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html#storage-inplace-encoding))
I was specifically reading this when thinking whether it's ok to keep them as an array in the struct

in any case I don't mind changing to two separate variables, just curious how the storage layout works

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: pushed a fix

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
eladmallel: @davidbrai I read this quote as saying that fixed-size arrays still start a new slot:

> Structs and array data always start a new slot and their items are packed tightly according to these rules.

(dynamic arrays are discussed in a different section just under)

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: subgraph updated

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
solimander: I thought the same, but I tested this in remix due to this line:

> Structs and array data always start a new slot and their items are packed tightly according to these rules.

Even with the optimizer on, it takes up two slots. You can see the same if you use `hardhat-storage-layout`.

davidbrai: @eladmallel ok I see. I misunderstood it then. thanks!

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+
+        _writeQuorumParamsCheckpoint(params);
+
+        emit DynamicQuorumParamsSet(
+            params.minQuorumVotesBPS,
+            params.maxQuorumVotesBPS,
+            params.quorumVotesBPSOffset,
+            params.quorumLinearCoef,
+            params.quorumQuadraticCoef
+        );
```
solimander, line 687: Thinking we could just reuse the above events for the aggregate function. Nice to only have a single event for the update of each dynamic quorum parameter.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+
+        uint16 oldQuorumVotesBPSOffset = params.quorumVotesBPSOffset;
+        params.quorumVotesBPSOffset = newQuorumVotesBPSOffset;
+
+        _writeQuorumParamsCheckpoint(params);
+
+        emit QuorumVotesBPSOffsetSet(oldQuorumVotesBPSOffset, newQuorumVotesBPSOffset);
+    }
+
+    function _setQuorumLinearCoef(uint32 newQuorumLinearCoef) external {
```
solimander, line 639: Nit: Do not abbreviate `coefficient`

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+     * @notice Admin function for setting the proposal threshold basis points
+     * @dev newProposalThresholdBPS must be greater than the hardcoded min
+     * @param newProposalThresholdBPS new proposal threshold
+     */
+    function _setProposalThresholdBPS(uint256 newProposalThresholdBPS) external {
+        require(msg.sender == admin, 'NounsDAO::_setProposalThresholdBPS: admin only');
+        require(
+            newProposalThresholdBPS >= MIN_PROPOSAL_THRESHOLD_BPS &&
+                newProposalThresholdBPS <= MAX_PROPOSAL_THRESHOLD_BPS,
+            'NounsDAO::_setProposalThreshold: invalid proposal threshold'
```
solimander, line 576: Leftover from V1, but could you update to `NounsDAO:: _setProposalThresholdBPS: invalid proposal threshold bps`

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+
+        uint32 oldQuorumQuadraticCoef = params.quorumQuadraticCoef;
+        params.quorumQuadraticCoef = newQuorumQuadraticCoef;
+
+        _writeQuorumParamsCheckpoint(params);
+
+        emit QuorumQuadraticCoefSet(oldQuorumQuadraticCoef, newQuorumQuadraticCoef);
+    }
+
+    function _setDynamicQuorumParams(DynamicQuorumParams memory params) public {
```
solimander, line 663: Should be able to add function comments to these now. Public interfaces look good.

packages/nouns-subgraph/README.md
```diff
@@ -37,3 +37,36 @@ yarn codegen
 ```sh
 yarn deploy:[network] # Supports rinkeby and mainnet
 ```
+
+## Running a local deployment
```
solimander, line 41: Worth mentioning that the local chain must be started by running `yarn task:run-local` from the contracts directory.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+        Proposal storage proposal = _proposals[proposalId];
+        if (proposal.totalSupply == 0) {
+            return proposal.minQuorumVotes;
+        }
+
+        return
+            dynamicQuorumVotes(
+                proposal.againstVotes,
+                proposal.totalSupply,
+                getDynamicQuorumParamsAt(proposalCreationBlock(proposal))
```
solimander, line 783: You can use `proposal.creationBlock` here as `proposal.creationBlock == 0` will never be true.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+     *       where coefs is `params.quorumPolynomCoefs`
+     * @dev Note that the quorumPolynomCoefs are fixed point integers with 6 decimals
+     * @param againstVotes Number of against-votes in the proposal
+     * @param totalSupply The total supply of Nouns at the time of proposal creation
+     * @return quorumVotes The required quorum
+     */
+    function dynamicQuorumVotes(
+        uint256 againstVotes,
+        uint256 totalSupply,
+        DynamicQuorumParams memory params
```
solimander, line 803: Add param comment for this one.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+    ) public pure returns (uint256) {
+        uint256 againstVotesBPS = (10000 * againstVotes) / totalSupply;
+        if (againstVotesBPS <= params.quorumVotesBPSOffset) {
+            return bps2Uint(params.minQuorumVotesBPS, totalSupply);
+        }
+
+        uint256 polynomInput = againstVotesBPS - params.quorumVotesBPSOffset;
```
solimander, line 810: Nit: Use more descriptive variable name. e.g. `adjustedAgainstVotesBPS`

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,895 @@
...
+        uint256 polynomValueBPS = (params.quorumLinearCoef *
```
solimander, line 811: Nit: Use more descriptive variable name. e.g. `quorumAdjustmentBPS`

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: fixed https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/5822d9fec5966061fd45e45f1a0c180466c2173d

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/605e2f61d272508affc85ecc33a8d2b28ee14295

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
```diff
@@ -214,6 +243,179 @@ contract NounsDAOStorageV1 is NounsDAOProxyStorage {
...
+
+    struct Proposal {
+        /// @notice Unique id for looking up a proposal
+        uint256 id;
+        /// @notice Creator of the proposal
+        address proposer;
+        /// @notice The number of votes needed to create a proposal at the time of proposal creation. *DIFFERS from GovernerBravo
+        uint256 proposalThreshold;
+        /// @notice The minimum number of votes in support of a proposal required in order for a quorum to be reached and for a vote to succeed at the time of proposal creation. *DIFFERS from GovernerBravo
+        uint256 minQuorumVotes;
```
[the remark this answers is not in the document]
davidbrai: fixed https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/19485af693d754736f92205085a8a6b6f1e4d8e4

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: nice catch!
fixed: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/38c4cc56b60954e0e32d50b2c4acebdc4717b75d

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/807c3adb12ee7aa08c43d41dee6807a8dd53645f
If you think the description of the `DynamicQuorumParams` struct members should also go here, lmk

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/f9dea74a67e70853d622a649bdffe895b68f86b7

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/f9dea74a67e70853d622a649bdffe895b68f86b7

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: added: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/951d57faa1096fdc82df3d7261fdc4d98570e3f3

packages/nouns-subgraph/README.md
eladmallel: fixed

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,966 @@
...
+            oldParams = getDynamicQuorumParamsAt(block.number);
+        } else {
+            oldParams = DynamicQuorumParams({
+                minQuorumVotesBPS: 0,
+                maxQuorumVotesBPS: 0,
+                quorumVotesBPSOffset: 0,
+                quorumLinearCoefficient: 0,
+                quorumQuadraticCoefficient: 0
+            });
+        }
```
solimander, line 741: Unnecessary `else` block. The default value of each of these is 0.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,966 @@
...
+            params.minQuorumVotesBPS > MIN_QUORUM_VOTES_BPS_UPPER_BOUND
+        ) {
+            revert InvalidMinQuorumVotesBPS();
+        }
+        if (params.maxQuorumVotesBPS > MAX_QUORUM_VOTES_BPS_UPPER_BOUND) {
+            revert InvalidMaxQuorumVotesBPS();
+        }
+        if (params.minQuorumVotesBPS > params.maxQuorumVotesBPS) {
+            revert MinQuorumBPSGreaterThanMaxQuorumBPS();
+        }
```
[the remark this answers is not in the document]
solimander: Ah, I see the comment related to contract size. This feels like a good opportunity to modernize and move this contract over to custom errors. Interested in thoughts from Max on this.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,966 @@
...
+        Proposal storage proposal = _proposals[proposalId];
+        if (proposal.totalSupply == 0) {
+            return proposal.quorumVotes;
+        }
+
+        return
+            dynamicQuorumVotes(
+                proposal.againstVotes,
+                proposal.totalSupply,
+                getDynamicQuorumParamsAt(proposal.creationBlock)
```
solimander, line 845: It's a bit scary that failure to set the dynamic quorum params in the same transaction as the V1 -> V2 replacement will bork governance for good. Let's take a look and see if there are any simple solutions to avoid this issue.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,966 @@
...
+
+    /**
+     * @notice Admin function for setting the minimum quorum votes bps
+     * @param newMinQuorumVotesBPS minimum quorum votes bps
+     *     Must be between `MIN_QUORUM_VOTES_BPS_LOWER_BOUND` and `MIN_QUORUM_VOTES_BPS_UPPER_BOUND`
+     *     Must be lower than or equal to maxQuorumVotesBPS
+     */
+    function _setMinQuorumVotesBPS(uint16 newMinQuorumVotesBPS) external {
+        require(msg.sender == admin, 'NounsDAO::_setMinQuorumVotesBPS: admin only');
+        DynamicQuorumParams memory params = getDynamicQuorumParamsAt(block.number);
```
solimander, line 613: Looking at the logic here, all of these would revert when upgrading from V1 -> V2 because `len > 0` would fail.

solimander: Can you confirm this one? Should be caught by tests.

eladmallel: fixed in https://github.com/nounsDAO/nouns-monorepo/commit/11dc46ccf886767298934511db674350131ebf99

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: great feedback; fixed in this commit to return default values instead https://github.com/nounsDAO/nouns-monorepo/commit/11dc46ccf886767298934511db674350131ebf99

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: removed in https://github.com/nounsDAO/nouns-monorepo/commit/11dc46ccf886767298934511db674350131ebf99

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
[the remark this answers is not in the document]
eladmallel: yeah I don't love the inconsistency either and would love to switch to customer errors; we're just trying to keep the amount of changes to a minimum to make this upgrade smoother

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -0,0 +1,979 @@
...
+    uint256 public constant MAX_VOTING_DELAY = 40_320; // About 1 week
+
+    /// @notice The lower bound of minimum quorum votes basis points
+    uint256 public constant MIN_QUORUM_VOTES_BPS_LOWER_BOUND = 200; // 200 basis points or 2%
+
+    /// @notice The upper bound of minimum quorum votes basis points
+    uint256 public constant MIN_QUORUM_VOTES_BPS_UPPER_BOUND = 2_000; // 2,000 basis points or 20%
+
+    /// @notice The upper bound of maximum quorum votes basis points
+    uint256 public constant MAX_QUORUM_VOTES_BPS_UPPER_BOUND = 4_000; // 4,000 basis points or 40%
```
solimander, line 88: Forgot to mention, we can probably bump this up to 60%. Still low enough to prevent governance from being fully locked by a bad update and high enough that we'd likely never want to exceed it.

eladmallel: fixed https://github.com/nounsDAO/nouns-monorepo/commit/01d3c306bb5f47af7e5970cf0d79be55f9b0b54e

packages/nouns-contracts/test/governance/NounsDAO/V2/dynamicQuorum.test.ts
```diff
@@ -27,96 +26,41 @@ describe('Dynamic Quorum', () => {
...
-        quorumLinearCoefficient: parseUnits('1', 6),
-        quorumQuadraticCoefficient: 0,
-      });
-
-      expect(quorumVotes).to.equal(20);
+  it('increases linearly', async () => {
+    const quorumVotes = await gov.dynamicQuorumVotes(18, 200, {
+      minQuorumVotesBPS: 1000,
+      maxQuorumVotesBPS: 4000,
+      quorumCoefficient: parseUnits('1', 6),
```
davidbrai, line 43: @eladmallel can you add one or two more tests with coefficients other 1?
e.g. 0.1 and 2

packages/nouns-contracts/test/governance/NounsDAO/V2/dynamicQuorum.test.ts
```diff
@@ -27,96 +26,41 @@ describe('Dynamic Quorum', () => {
...
-        minQuorumVotesBPS: 1000,
-        maxQuorumVotesBPS: 4000,
-        quorumVotesBPSOffset: 500,
-        quorumLinearCoefficient: parseUnits('1', 6),
-        quorumQuadraticCoefficient: 0,
-      });
-
-      expect(quorumVotes).to.equal(28);
-    });
+    expect(quorumVotes).to.equal(38);
```
davidbrai, line 46: @eladmallel also to test the "linearity", could you add a few more checks at points other than 18 against votes?

packages/nouns-contracts/test/governance/NounsDAO/V2/quorumConfig.test.ts
```diff
@@ -75,102 +75,73 @@ describe('NounsDAOV2#_setDynamicQuorumParams', () => {
...
-      .to.emit(gov, 'QuorumQuadraticCoefficientSet')
-      .withArgs(parseUnits('2', 6), quorumQuadraticCoefficient);
+      .to.emit(gov, 'QuorumCoefficientSet')
+      .withArgs(parseUnits('1', 6), quorumCoefficient);
   });
 
   it('given no prior config, sets value and emits event', async () => {
-    const quorumLinearCoefficient = parseUnits('1', 6);
+    const quorumCoefficient = parseUnits('1', 6);
     const quorumQuadraticCoefficient = parseUnits('0.001', 6);
```
davidbrai, line 134: @eladmallel leftover

packages/nouns-contracts/test/governance/NounsDAO/V2/dynamicQuorum.test.ts
eladmallel: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/bd4499a65215dc671704b494ca53ab51d3e4e0e4

packages/nouns-contracts/test/governance/NounsDAO/V2/dynamicQuorum.test.ts
eladmallel: https://github.com/nounsDAO/nouns-monorepo/pull/438/commits/bd4499a65215dc671704b494ca53ab51d3e4e0e4

packages/nouns-contracts/tasks/create-proposal.ts
```diff
@@ -9,7 +9,7 @@ task('create-proposal', 'Create a governance proposal')
     types.string,
   )
   .setAction(async ({ nounsDaoProxy }, { ethers }) => {
-    const nounsDaoFactory = await ethers.getContractFactory('NounsDAOLogicV1');
+    const nounsDaoFactory = await ethers.getContractFactory('NounsDAOLogicV2');
```
davidbrai, line 12: @eladmallel should this change back to V1 until we upgrade it?

packages/nouns-contracts/tasks/deploy-local.ts
```diff
@@ -105,22 +117,28 @@ task('deploy-local', 'Deploy contracts to hardhat')
       NounsDAOExecutor: {
         args: [expectedNounsDAOProxyAddress, args.timelockDelay],
       },
-      NounsDAOLogicV1: {
+      NounsDAOLogicV2: {
         waitForConfirmation: true,
       },
-      NounsDAOProxy: {
+      NounsDAOProxyV2: {
```
davidbrai, line 123: @eladmallel same as above, should this stay V1 until we upgrade?

eladmallel: in this case I'd rather keep it on v2, since we're very close to deploying it, and we're not really relying on local runs to test the upgrade process. wdyt?

packages/nouns-contracts/tasks/create-proposal.ts
eladmallel: in practice I think this doesn't matter much, because the propose function signature is the same, but I think it's easy to change :)

packages/nouns-contracts/tasks/deploy-local.ts
davidbrai: I don't know when we'll actually do the upgrade.
I think keep it on V1 until then can be less confusing for potential forkers

solimander (approved):

PR #442, opened 2022-05-10
Add proposer "by line" and "ENS everywhere" to prop page

bcjgit:
# TLDR
This proposal does 2.5 things:
1. Ensures all addresses in proposal transactions resolve ENS if one exists
2. Replaces the `Proposer` section at the bottom of the props with a `Proposed by` "by line"
3. Makes sure the "by line" is translated into Japanese and specific formatting is applied in this case (to account for the different ordering of words between `en-US` and `ja-JP`

# Testing done
1. Tested on mobile (iPhone, chrome dev tools) and desktop (chrome)
2. Confirmed translations make sense with translator

# Open questions

One thing that I don't *love* about this PR is that I special case Japanese. I couldn't find a way around it in this case but don't love setting that precedent. Any thoughts on how to avoid this would be welcomed.


packages/nouns-webapp/src/components/ProposalHeader/index.tsx
```diff
@@ -108,6 +134,31 @@ const ProposalHeader: React.FC<ProposalHeaderProps> = props => {
...
+              <span className={classes.proposedByJp}>Proposed by: </span>
+              <span className={classes.proposerJp}>
+              {proposer}
+              </span>
+              <span className={classes.propTransactionWrapperJp}>{proposedAtTransactionHash}</span>
+            </Trans>
+          </div>
+        ) : (
+          <>
+            <h3>Proposed by:</h3>
```
solimander, line 150: super-nit: cleaner without the colon imo

solimander (approved):
Looks good pending merge conflict fixes

PR #444, opened 2022-05-10
Add proposal state transition countdown (Starting in, Ending in, Expiring in)

bcjgit:


solimander (approved):
Really helpful addition 👍

PR #447, opened 2022-05-16
Add chunking to logs call

bcjgit:


solimander:
Implemented in https://github.com/nounsDAO/nouns-monorepo/pull/448

PR #450, opened 2022-05-17
Use Read-Only Provider to Fetch Logs

solimander:
Force `getLogs` to use the readonly provider, rather than the wallet provider. The wallet provider (Infura) times out far faster than Alchemy.


gitblamecarrot (approved):
Should be OK as a hotfix. My concern is it'll remove the ability to use an injected provider in case Alchemy is having troubles. We're seeing the opposite now but we should circle back and reevaluate at some point if we can go back to injected.

PR #460, opened 2022-05-23
Frontend Optimizations (Caching & Subgraph Consumption)

solimander:
This PR contains several optimizations, which speed up the loading of Nouns and proposals.

1. Cache Noun seeds.
2. Fetch proposals from the subgraph, falling back to the chain if necessary.

### DIFF
--- a/packages/nouns-webapp/src/wrappers/nounsDao.ts
+++ b/packages/nouns-webapp/src/wrappers/nounsDao.ts
@@ -1,6 +1,8 @@
 import { NounsDAOABI, NounsDaoLogicV1Factory } from '@nouns/sdk';
 import {
   ChainId,
+  useBlockMeta,
+  useBlockNumber,
   useContractCall,
   useContractCalls,
   useContractFunction,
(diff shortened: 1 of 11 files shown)


packages/nouns-webapp/src/wrappers/nounsDao.ts
```diff
@@ -182,67 +244,144 @@ const countToIndices = (count: number | undefined) => {
...
+        return ProposalState.SUCCEEDED;
+      }
+    }
+    return status;
+  }
+  if (status === ProposalState.QUEUED) {
+    if (!blockTimestamp || !proposal.executionETA) {
+      return ProposalState.UNDETERMINED;
+    }
+    const GRACE_PERIOD = 14 * 60 * 60 * 24;
```
bcjgit, line 320: nit: This could change if we were to upgrade to gov v3 right? Probably fine just flagging

bcjgit (approved):
Left 1 small non blocking question nit 

overall LGTM (admittedly, didn't like go into the specifics of the regexes or anything) but overall logic seemed reasonable to me.

packages/nouns-webapp/src/wrappers/nounsDao.ts
solimander: Could potentially be swapped out someday, but very unlikely because it would require swapping out the executor and this value seems reasonable.

PR #464, opened 2022-05-25
Dependency Upgrades

solimander:
* Upgrade `ethers` to fix a bug in `bn.js`
* Upgrade `@davatar/react` to reduce network calls
* All other upgrades are required by the new `ethers` or `@davatar/react` versions


bcjgit:
I think you need to change this to a non draft review for me to formally ship but per our discussion earlier today LGTM

bcjgit (approved):
LGTM

PR #468, opened 2022-05-30
Delegation modal

bcjgit:
# TLDR

This PR adds a Modal to to make it easy to delegate votes. Button appears for users with Nouns on the `/vote` page. 

# Testing done
- Testing on desktop and mobile (both en and ja) and it looked fine
- Tested delegation on Rinkeby and it seemed to work


solimander:
Starting review in the morning 🌞

packages/nouns-webapp/src/components/ChangeDelegatePannel/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+      ChangeDelegateState.CHANGE_FAILURE,
+    ],
+    [
+      <NavBarButton
+        buttonText={
+          <div className={classes.delegateKVotesBtn}>
+            {locale === 'en-US ' ? (
+              <>
+                Delegate <span className={classes.highlightCircle}>{availableVotes}</span>
+                {availableVotes === 1 ? <>Votes</> : <>Votes</>}
```
solimander, line 115: Should be `{availableVotes === 1 ? <>Vote</> : <>Votes</>}`

packages/nouns-webapp/src/components/ChangeDelegatePannel/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+      </Trans>,
+      <Trans>
+        Your <span style={{ fontWeight: 'bold' }}>{availableVotes}</span> votes are being delegated
+        to a new account.
+      </Trans>,
+      <Trans>
+        Your <span style={{ fontWeight: 'bold' }}>{availableVotes}</span> votes have been delegated
+        to a new account.
+      </Trans>,
+      ,
```
solimander, line 180: nit: remove extra comma

packages/nouns-webapp/src/components/Proposals/index.tsx
```diff
@@ -109,29 +129,19 @@ const Proposals = ({ proposals }: { proposals: Proposal[] }) => {
...
-                      <div className={classes.countdownPillContentWrapper}>
-                        <span className={classes.countdownPillClock}>
-                          <ClockIcon height={16} width={16} />
-                        </span>{' '}
-                        <span className={classes.countdownPillText}>
-                          {getCountdownCopy(p, currentBlock || 0, activeLocale)}
-                        </span>
-                      </div>
-                    </div>
+            const coundownPill = (
```
solimander, line 132: nit: `coundownPill` -> `countdownPill`

packages/nouns-webapp/src/wrappers/nounToken.ts
```diff
@@ -146,3 +154,22 @@ export const useUserVotesAsOfBlock = (block: number | undefined): number | undef
     }) || [];
   return votes?.toNumber();
 };
+
+export const useDelegateVotes = (delagee: string) => {
```
solimander, line 158: Remove the unused param

packages/nouns-webapp/src/components/ChangeDelegatePannel/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+    ChangeDelegateState.ENTER_DELEGATE_ADDRESS,
+  );
+
+  const { library, account } = useEthers();
+
+  const [delegateAddress, setDelegateAddress] = useState('');
+  const [delegateInputText, setDelegateInputText] = useState('');
+  const [delegateInputClass, setDelegateInputClass] = useState<string>('');
+  const availableVotes = useNounTokenBalance(account ?? '') ?? 0;
+  const { send: delegateVotes, state: delageeState } = useDelegateVotes(delegateAddress);
```
solimander, line 57: `delageeState` -> `delegateState` or `delegateeState`

solimander:
Hitting the point where we should dig into the network tab to see why so many external calls are being made. Load the proposals page, open the delegation modal, click 'Update Delegate' and we're already at 220 calls. Many of them are RPC calls.

bcjgit:
> 

I think this might be independent of the modal. If I comment out the modal and load the page for an account with Nouns in it (rinkeby) I see ~300 requests after about a minute of waiting. When I reload the page with the modal I see ~ the same number (actually a bit less) and hitting the update button is only netting an incremental O(10s) of requests. 

Could be wrong but my hunch is this might not be (at least directly) related to the modal. 

--
FWIW: Even for an account without Nouns on nouns.wtf/vote I see ~150 requests many of which are RPC calls ... would be curious how this changes if the connected account holds Nouns as I found on rinkeby that when I didn't hold Nouns the number of calls was also about 100-150.

solimander:
> I think this might be independent of the modal. If I comment out the modal and load the page for an account with Nouns in it (rinkeby) I see ~300 requests after about a minute of waiting. When I reload the page with the modal I see ~ the same number (actually a bit less) and hitting the update button is only netting an incremental O(10s) of requests.

> Could be wrong but my hunch is this might not be (at least directly) related to the modal.

Oof, okay. I was seeing an immediate 20 request increase when opening the modal, but it sounds like it's a symptom of another issue. Let's create a separate issue to diagnose and address the large number of outgoing requests.

solimander (approved):
Nice work! Not high priority, but may be worth adding a tooltip to notify the delegator that this will only change the voting power on proposals that haven't been created yet (i.e. no effect on pending proposals).

PR #470, opened 2022-05-31
swap winner and holder in profile UI

bcjgit:
**TLDR:**
per https://discord.com/channels/849745721544146955/873197966373052426/980917083493773334 Swaps Winner and holder in profiles UI

Testing done:

Looked at profile (nounder and non-nounder) and it seems to work correctly.

Will also watch today's auction to make sure nothing looks strange.

**Notes to reviewer:**
One thing that could be helpful to get a double check on is confirming that in the case of an auction being won we will default to the `Winner` component vs the holder component.

### DIFF
--- a/packages/nouns-webapp/src/locales/en-US.po
+++ b/packages/nouns-webapp/src/locales/en-US.po
@@ -396,10 +396,6 @@ msgstr "Held by"
 msgid "Help mint the next Noun"
 msgstr "Help mint the next Noun"
 
-#: src/components/Holder/index.tsx
-#~ msgid "Holder"
-#~ msgstr "Holder"
-
 #: src/components/Documentation/index.tsx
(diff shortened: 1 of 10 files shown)


packages/nouns-webapp/src/locales/en-US.po
```diff
@@ -387,14 +388,18 @@ msgstr "Governance ‘Slow Start’"
...
 msgid "Held by"
 msgstr "Held by"
 
 #: src/components/AuctionActivity/index.tsx
 msgid "Help mint the next Noun"
 msgstr "Help mint the next Noun"
 
+#: src/components/Holder/index.tsx
+#~ msgid "Holder"
+#~ msgstr "Holder"
```
solimander, line 401: Can this be removed from the language files?

solimander (approved):

PR #471, opened 2022-05-31
filter pre vote canceled props from noun profiles

bcjgit:
TLDR:

re https://discord.com/channels/849745721544146955/877215225093451776/977694339176214529

This PR filters props that are cancelled and got 0 votes of any kind (for, against, abstain) from the profile activity feed. 

Testing done:

Confirmed by looking at profile activity feed that props cancelled before going to a vote were filtered

### DIFF
--- a/packages/nouns-webapp/src/components/ProfileActivityFeed/index.tsx
+++ b/packages/nouns-webapp/src/components/ProfileActivityFeed/index.tsx
@@ -65,15 +65,22 @@ const ProfileActivityFeed: React.FC<ProfileActivityFeedProps> = props => {
     }, {});
 
   const filteredProposals = proposals.filter((p: Proposal, id: number) => {
-    return (
-      // Filter props from before the Noun was born
-      (parseInt(proposalCreatedTimestamps.proposals[id].createdTimestamp) >
-        nounCanVoteTimestamp.toNumber() ||
-        (p.id && nounVotes[p.id])) &&
-      // Filter props which were cancelled and got 0 votes of any kind
-      // This a hack to filter props that were cancelled before going live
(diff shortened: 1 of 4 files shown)


packages/nouns-webapp/src/components/ProfileActivityFeed/index.tsx
```diff
@@ -66,9 +66,13 @@ const ProfileActivityFeed: React.FC<ProfileActivityFeedProps> = props => {
...
+      // Filter props from before the Noun was born
+      (parseInt(proposalCreatedTimestamps.proposals[id].createdTimestamp) >
         nounCanVoteTimestamp.toNumber() ||
-      (p.id && nounVotes[p.id])
+        (p.id && nounVotes[p.id])) &&
+      // Filter props which were cancelled and got 0 votes of any kind
+      // This a hack to filter props that were cancelled before going live
+      !(p.status === ProposalState.CANCELLED && p.forCount + p.abstainCount + p.againstCount === 0)
     );
   });
```
solimander, line 77: nit: can make this a bit more readable -

```ts
const filteredProposals = proposals.filter((p: Proposal, id: number) => {
  const proposalCreationTimestamp = parseInt(proposalCreatedTimestamps.proposals[id].createdTimestamp);

  // Filter props from before the Noun was born
  if (nounCanVoteTimestamp.gt(proposalCreationTimestamp)) {
    return false;
  }
  // Filter props which were cancelled and got 0 votes of any kind
  if (p.status === ProposalState.CANCELLED && p.forCount + p.abstainCount + p.againstCount === 0) {
    return false;
  }
  return true;
});
```

solimander (approved):

PR #472, opened 2022-06-01
Foundry tests

davidbrai:
Add support for foundry tests.
This setup allows writing and running foundry tests in parallel to the hardhat setup.

### DIFF
--- a/packages/nouns-contracts/foundry.toml
+++ b/packages/nouns-contracts/foundry.toml
@@ -1,4 +1,5 @@
 [default]
 src = 'contracts'
 libs = ['lib', '../../node_modules']
-
+cache_path = 'foundry-cache'
+out = 'foundry-out'
--- a/packages/nouns-contracts/.gitignore
+++ b/packages/nouns-contracts/.gitignore
@@ -3,7 +3,6 @@ parts.html
 
 #Hardhat files
 cache
-hh-cache
 artifacts
 typechain
 .openzeppelin
@@ -22,3 +21,7 @@ abi/contracts/governance
 abi/contracts/interfaces
 abi/contracts/proxies
 abi/contracts/test
+
+# Foundry
+foundry-out
+foundry-cache
--- a/packages/nouns-contracts/hardhat.config.ts
+++ b/packages/nouns-contracts/hardhat.config.ts
@@ -57,7 +57,5 @@ const config: HardhatUserConfig = {
   mocha: {
     timeout: 60_000,
   },
-  // Avoid foundry cache conflict.
-  paths: { cache: 'hh-cache' },
 };
 export default config;


solimander:
Could you update the `nouns-contracts` README with instructions for Foundry, please? 🙏 

Should include `forge install`, which is required to install `forge-std`

solimander:
Should add the following to the top-level `.vscode/settings.json` so `forge-std` and `ds-test` resolve in VS Code:


```json
"solidity.remappings": [
  "ds-test/=packages/nouns-contracts/lib/ds-test/src/",
  "forge-std/=packages/nouns-contracts/lib/forge-std/src/"
],
```

davidbrai:
@solimander regarding the remappings:

every time we add a forge dependency we'll need to update it, usually I just do `forge remappings > remappings.txt`

vscode can use the remappings.txt file, so it works when opening vscode from the nouns-contracts folder.
in order to make it also work when opening from the root folder, perhaps we can just do a soft link named remappings.txt pointing to the remappings.txt inside nouns-contracts?

solimander:
> perhaps we can just do a soft link named remappings.txt pointing to the remappings.txt inside nouns-contracts?

Yeah, that'd be great if you're able to get it to work from both directories with a symlink.

davidbrai:
@solimander I pushed the symlinks. had to create a symlink for the lib folder also. lmk if it works for you also

davidbrai:
@solimander looks like CI is also working
could you please review and lmk if anything is missing?

solimander:
Awesome, pulling local now!

packages/nouns-contracts/foundry.toml
```diff
@@ -0,0 +1,4 @@
+[default]
+src = 'contracts'
+libs = ['lib', '../../node_modules']
+
```
solimander, line 4: nit: remove extra line

solimander:
> I pushed the symlinks. had to create a symlink for the lib folder also. lmk if it works for you also

Nice, works for me too

solimander:
Only issue I see is that `yarn build:sol` fails following a `forge build` due to typechain type issues.

davidbrai:
> Only issue I see is that `yarn build:sol` fails following a `forge build` due to typechain type issues.

I'm not getting any issue.
I cloned a fresh copy and ran these commands:

```
git clone git@github.com:nounsDAO/nouns-monorepo.git foundry-tests
cd foundry-tests
git checkout davidbrai-foundry-tests
yarn
cd packages/nouns-contracts
forge install
forge build
yarn build:sol
yarn build:sol --force
```

all of them worked.

how can I reproduce your issue?

solimander:
Ah, the issue comes when running `yarn build`, rather than `yarn build:sol`

davidbrai:
> Ah, the issue comes when running `yarn build`, rather than `yarn build:sol`

This was a really nasty one. I tracked it down to be caused by the cache directory. Even though I pointed hardhat to a separate one, I think the typechain plugin was looking at `cache`.
Changing hardhat to use `cache` and foundry to use a different one seems to solve it.
Will do some more testing and push

devanonon:
Tiny point, but I would recommend running foundry tests in another job, CI will complete faster as jobs run in parallel by default.

Example here: https://github.com/foundry-rs/hardhat-foundry-template/blob/master/.github/workflows/test.yml

davidbrai:
@solimander pushed a commit which I think fixes the build issue
would be great if you can make sure it doesn't for you either

solimander (approved):
🚀

PR #477, opened 2022-06-05
Fix image encoding for non-32px high images

c0:
- The logic for determining image bounds assumes the height of the image is 32px which results in incorrect bounds of `{ top: 1, bottom: 0, left: 0, right: 0 }`
- Addresses the issue by using the image height to determine the last row
- Adds a test case for an oddly sized image


volkyeth:
works as intended. code is meant for 32x32px imgs alone

PR #478, opened 2022-06-06
Bug Fix: Check Seed Validity Before Caching

solimander:
Sometimes, immediately following settlement, the `seed` contract call response is `[0, 0, 0, 0, 0]`. Previously, this invalid seed was cached in local storage.

To fix this issue, we now check the validity of the seed before caching.


bcjgit (approved):
🚢

PR #482, opened 2022-06-12
Full image support in encoder

isamu:
Thank you for the wonderful project!

I'm using nouns-sdk to create our nounish NFTs.

I prepared a 32x32 image as a background image and encoded it, but there was a problem that the left and right edges and the top were cut off.

Original asset
https://github.com/Cryptocoders-wtf/sushi-nouns-dao-assets/blob/main/images/0-backgrounds/bg-koushi.png

Generated NFT
https://testnets.opensea.io/assets/rinkeby/0xa87701042589902151d7fca744067db2443b0ce6/8

After investigating, it seems that nouns usually treats the left and right colors as margins because the picture is not drawn on the entire 32x32 image.
As a workaround for this, pass a flag so that all pixels are run-length coded as an image.

Also, there is another problem.
There is a problem that the first row y = 0 image is not handled.
I can't tell if this is a bug or a spec.
The initial value of top of boundary is set to -1 instead of 0, and it is processed correctly when there is data at row y = 0.

With these changes, we have confirmed that our background image is encoded correctly.

https://testnets.opensea.io/assets/rinkeby/0x5b4691811c34ac17f39b0a11b66ed25f73953fe3/23

We also coded the existing nouns assets and confirmed that our changes do not affect the coded data of the existing data.


volkyeth:
Nouns contracts do not support background images. The backgrounds are just plain colors.

PR #483, opened 2022-06-12
i18n: Brazilian Portuguese - pt-BR by hOurNouns - cherrry picked from…

eduardocruz:
… https://github.com/hOurNouns/hournouns-monorepo/tree/feature/i18n_pt_br

### DIFF
--- a/packages/nouns-webapp/src/locales/pt-BR.po
+++ b/packages/nouns-webapp/src/locales/pt-BR.po
@@ -33,15 +33,15 @@ msgstr ""
 
 #: src/components/VoteCard/index.tsx
 msgid "Abstain"
-msgstr "Abstenções"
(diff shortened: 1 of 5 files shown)


packages/nouns-webapp/src/locales/pt-BR.po
danicuki, line 1: tem o arquivo `pt-BR.js` que também precisa ser criado e traduzido, similar ao `ja-JP.js`

danicuki:
This is a pretty old PR that should be taken into consideration. Brazilian Nouns community is huge and this PR can make it even more big and relevant.

volkyeth:
This is pretty stale, so I'll close it. Would consider merging an up-to-date one.

Let's make nouns.wtf global

PR #515, opened 2022-07-14
Voting refund

davidbrai:
This PR introduces new functions that allow token holders to vote and receive a gas refund from the DAO in the same transaction.

For a more detailed spec see: https://github.com/nounsDAO/nouns-tech/tree/main/vote-refund


packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -901,6 +964,14 @@ contract NounsDAOLogicV2 is NounsDAOStorageV2, NounsDAOEventsV2 {
         }
     }
 
+    function _refundGas(uint256 startGas) internal {
+        uint256 gasPrice = min(tx.gasprice, block.basefee + MAX_REFUND_PRIORITY_FEE);
+        uint256 gasUsed = startGas - gasleft() + REFUND_TRANSFER_GAS + REFUND_OVERHEAD_GAS;
```
davidbrai, line 969: @eladmallel maybe worth merging REFUND_TRANSFER_GAS and REFUND_OVERHEAD_GAS to one constant.
in case the compiler doesn't do so already, it's not worth the extra addition operation.
you can specify the breakdown in a comment

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -484,6 +493,49 @@ contract NounsDAOLogicV2 is NounsDAOStorageV2, NounsDAOEventsV2 {
...
+     * @param support The support value for the vote. 0=against, 1=for, 2=abstain
+     * @param reason The reason given for the vote by the voter
+     * @dev Reentrancy is defended against in `castVoteInternal` at the `receipt.hasVoted == false` require statement.
+     */
+    function castRefundableVoteWithReason(
+        uint256 proposalId,
+        uint8 support,
+        string calldata reason
+    ) external {
+        uint256 startGas = gasleft();
```
davidbrai, line 531: minor: can extract all the code here and above to a `castRefundableVoteInternal`. it's almost the same code, just the `reason` differs

packages/nouns-contracts/test/governance/NounsDAO/V2/voteRefund.test.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  beforeEach(async () => {
+    snapshotId = await ethers.provider.send('evm_snapshot', []);
+  });
+
+  afterEach(async () => {
+    await ethers.provider.send('evm_revert', [snapshotId]);
+  });
+
+  describe.only('withdraw', () => {
```
davidbrai, line 63: @eladmallel leftover `.only`

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -901,6 +964,14 @@ contract NounsDAOLogicV2 is NounsDAOStorageV2, NounsDAOEventsV2 {
         }
     }
 
+    function _refundGas(uint256 startGas) internal {
```
davidbrai, line 967: @eladmallel maybe we can do an early exit here in case the contract's balance is 0?
no reason to waste user's gas on this

eladmallel: that's exactly what I was thinking -- there will be balance most of the time, so I didn't want to waste gas on checking, that's why it's implemented as it is. ok to keep it this way?

davidbrai: yeah

solimander:
How are we looking on the size of the `NounsDAOLogicV2` bytecode?

eladmallel:
> How are we looking on the size of the `NounsDAOLogicV2` bytecode?

it's deploying locally without warnings; do you want to know the actual contract size we're getting?

solimander:
> it's deploying locally without warnings; do you want to know the actual contract size we're getting?

Great, no worries then. Wasn't sure if we were really close bumping up against the limit.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -901,6 +971,14 @@ contract NounsDAOLogicV2 is NounsDAOStorageV2, NounsDAOEventsV2 {
...
+        uint256 gasPrice = min(tx.gasprice, block.basefee + MAX_REFUND_PRIORITY_FEE);
+        uint256 gasUsed = startGas - gasleft() + REFUND_BASE_GAS;
+        uint256 refundAmount = min(gasPrice * gasUsed, address(this).balance);
+        (bool refundSent, ) = msg.sender.call{ value: refundAmount }('');
+        emit RefundableVote(msg.sender, refundAmount, refundSent);
```
solimander, line 979: Looking at the calculations in this block, overflow should not be possible. How about wrapping in `unchecked` to save a few hundred gas?

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: agreed, fixed https://github.com/nounsDAO/nouns-monorepo/pull/515/commits/69350411bfe1d6b34850ccf287e34a1881f1364c

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: absolutely: https://github.com/nounsDAO/nouns-monorepo/pull/515/commits/2a911cf693a4c99c7aedb2c4061389bfc9d5f992

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -91,6 +91,12 @@ contract NounsDAOLogicV2 is NounsDAOStorageV2, NounsDAOEventsV2 {
     /// @notice The maximum number of actions that can be included in a proposal
     uint256 public constant proposalMaxOperations = 10; // 10 actions
 
+    /// @notice The maximum priority fee used to cap gas refunds in `castRefundableVote`
+    uint256 public constant MAX_REFUND_PRIORITY_FEE = 20 gwei;
```
solimander, line 95: A 20 gwei priority fee feels a bit excessive. How do you feel about reducing to 10 gwei?

eladmallel: was waiting to discuss this with you :)
the reason I went with 20 rather than a lower number was because I was anchored by these two implementations:

*  https://github.com/fomo-nouns/fomo-nouns/blob/main/packages/contracts/contracts/NounSettlement.sol#L17
* https://github.com/z0r0z/zolidity/blob/main/src/utils/Refunded.sol#L18

I'm really not sure why they chose such a high number.

Curious how did you land on 10 gwei? @solimander

solimander: The latter is closer to what we want. The former (FOMO) is incredibly high because they're trying to get into the pending block. Either way, lower stakes here because the transaction will not revert if exceeded.

Thinking through it again, even 10 is very high haha. In practice, the priority fee is unlikely to exceed 2 gwei... so maybe cap somewhere in the 2-5 range.

eladmallel: updated to 2 gwei

PR #548, opened 2022-08-18
 Merge PR from NNS team to support NNS + ENS in nouns.wtf

bcjgit:
# TLDR

PR to integrate NNS (Nounish ENS) into nouns.wtf. This pr is essentially (https://github.com/nounsDAO/nouns-monorepo/pull/547) contributed by the NNS team with a few minor modifications

# Testing done

- looked at a bunch of auction pages on mobile and desktop (hardcoding NNS enabled addresses in various places) and it looked fine
- Looked at prop pages at prop with NNS enabled proposers / voters and they looked fine on mobile and desktop

# Notes to reviewers  (:sparkles:  important - potential security concerns :sparkles: )

Only thing on my mind to note is that we are using the contract provided by the NNS team to now resolve NNS and ENS in a single contract call ... is this something we want to do? Could there be long term concerns with this approach (from a security point of view in particular)? 

If we don't want to do this we can instead hit the NNS contract and then fallback to (with a second contract call) hitting the ENS contract ourselves, at the cost of doubling the number of calls we need to resolve ENS (granted we cache these calls so maybe it's not a _huge_ deal to do that).


vapeape4464 (approved):
call looks harmless (view function & static pointers to NNS/ENS) especially if we have it hardcoded. primary concern is NNS is taken over and we resolve/display some bad names temporarily perhaps but easy to remove if needed

PR #559, opened 2022-09-01
Vote refund: add caps

eladmallel:


packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -97,6 +97,9 @@ contract NounsDAOLogicV2 is NounsDAOStorageV2, NounsDAOEventsV2 {
     /// @notice The vote refund gas overhead, including 7K for ETH transfer and 29K for general transaction overhead
     uint256 public constant REFUND_BASE_GAS = 36000;
 
+    /// @notice The maximum gas units the DAO will refund voters on
+    uint256 public constant MAX_REFUND_GAS_USED = 200_000;
```
davidbrai, line 101: can you add to the documentation how you decided on the value of 200K?
how long of a reason text does that cover?

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
```diff
@@ -100,6 +100,9 @@ contract NounsDAOLogicV2 is NounsDAOStorageV2, NounsDAOEventsV2 {
     /// @notice The maximum gas units the DAO will refund voters on
     uint256 public constant MAX_REFUND_GAS_USED = 200_000;
 
+    /// @notice The maximum basefee the DAO will refund voters on
+    uint256 public constant MAX_REFUND_BASE_FEE = 50 gwei;
```
davidbrai, line 104: how did you decide on 50 gwei?
seems pretty low.
I forked some query on dune here (extended the time range): https://dune.com/queries/1233037
IIUC it shows daily medians.
maybe set the max at 400 or 500 gwei?

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: it covers the reason text in the tests, which is 2 times the longest reason I found (from honk).
decided on this value by seeing how much that text got me to (about 220K gas) then cut back a bit.

honk's reason was 882 words, so this supports about 1764 words.
I'm thinking reason strings shouldn't be extremely long (as opposed to proposal descriptions that can get quite long)

wdyt?

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
davidbrai: the gas refund becomes more beneficial as the blockchain becomes more congested and the base fee rises.
I think making this value too low is worse than making it too high.
that's why I'm tending to be much more generous and go up to 400/500, planning for the next bull market

eladmallel: @solimander curious for your take please

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: I'll add to the natspec a specific number of words / characters this limit supports

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
solimander: 50 feels too low and 400/500 feels too high IMO. Looking at [data from the last bull market](https://dune.com/queries/1235429), we've sustained above 100 gwei for an entire voting period, but almost never 200 gwei for that amount of time. If past bull market gas prices are any indication of the future, a 200 gwei cap could make voters think twice about voting during a short-lived gas spike, while allowing for a full refund some time during the voting period.

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: fixed

packages/nouns-contracts/contracts/governance/NounsDAOLogicV2.sol
eladmallel: I've been reluctant to make this cap configurable since we already have a non-trivial number of nobs added in V2, but since we can't really predict gas usage future, perhaps we should reconsider changing these constant values to configurable state variables.

wdyt?

davidbrai: even if we leave it constant, since the gov contract is upgradeable, we could theoretically upgrade it just for changing the constant value. it's not the cleanest, but it's possible.

an advantage of keeping it constant is also reducing the gas cost of reading this value from storage.

I don't have a strong opinion here, but leaning towards setting a constant cap (not configurable).

solimander: I'm leaning towards keeping it constant as well. Doesn't feel like a compelling enough reason to add another state variable.

solimander (approved):

davidbrai:
related code4rena issue: https://github.com/code-423n4/2022-08-nounsdao-findings/issues/174

PR #562, opened 2022-09-10
fix cancel proposal requirement in V1 dao contract

davidbrai:
it should check that the proposal does not have more votes than the proposal threshold, i.e less than or equal to proposal threshold should allow canceling

issue from: https://github.com/code-423n4/2022-08-nounsdao-findings/issues/255


eladmallel (approved):
LGTM

PR #563, opened 2022-09-10
fix delegateBySig to not allow delegating to address(0)

davidbrai:
instead when address(0) is the delegatee it does self delegation, similar to the implementation in `delegate(address)`

this is a fix following c4 issue: https://github.com/code-423n4/2022-08-nounsdao-findings/issues/157


eladmallel (approved):
LGTM

davidbrai:
@solimander could you review please?
(this is a PR to be merged into `contract-fixes` branch, which contains fixes that can't/won't be deployed for nouns dao)

solimander (approved):

PR #575, opened 2022-09-24
Improve vote modal UI

bcjgit:


packages/nouns-webapp/package.json
```diff
@@ -55,7 +55,9 @@
     "react-stepz": "^1.0.2",
     "redux": "^4.1.0",
     "remark-breaks": "^3.0.1",
-    "web-vitals": "^1.0.1"
+    "web-vitals": "^1.0.1",
+    "react-transition-group": "^4.4.5",
+    "@types/react-transition-group": "^4.4.5"
```
solimander, line 60: nit: move `@types/react-transition-group` to `devDependencies `

packages/nouns-webapp/src/components/VoteModal/index.tsx
```diff
@@ -204,6 +201,18 @@ const VoteModal = ({ show, onHide, proposalId, availableVotes }: VoteModalProps)
...
           </Button>
+
+          <div
+          className={classes.gasFreeVotingWrapper}
+          >
+            <span
+            className={classes.gasFreeVotingCopy}
+            >
+              <Trans>
+              Gas spent on voting will be refunded to you
```
solimander, line 212: Worth mentioning that gas will be refunded to the direct caller in the current version? If voting from a Gnosis Safe, the refund will go to the safe.

PR #585, opened 2022-09-29
Update nouns.wtf for DAO v2

bcjgit:
# TLDR

Adds:

- support for Goreli 
- DAO v2
- Updated visualization of dynamic quorum 

## Testing Done

Tested the following on Goreli:
- submit, vote on and pass v1 props
- can upgrade to v2
- v1 props started before v2 upgrade but ending after v2 upgrade work as expected
- v2 props can submitted, voted on, and passed
- dynamic quorum visualization works as expected in all cases


packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -154,9 +153,18 @@ export function handleVoteCast(event: VoteCast): void {
...
+  } else if (event.params.support == 1) {
+    proposal.forVotes = proposal.forVotes.plus(event.params.votes);
+  } else if (event.params.support == 2) {
+    proposal.abstainVotes = proposal.abstainVotes.plus(event.params.votes);
   }
 
-  if (proposal.usingDynamicQuorum) {
+  const dqParams = getOrCreateDynamicQuorumParams();
+  const usingDynamicQuorum =
+    dqParams.dynamicQuorumStartBlock !== null &&
```
solimander, line 164: I don't believe that it matters when comparing against `null`, but it's best practice to use `==` and `!=` in AssemblyScript comparisons. `===` and `!==` compare references in AssemblyScript.

solimander: There are 3 instances of this in the subgraph that should be updated just in case.

solimander (approved):

PR #598, opened 2022-10-18
Auction House V2: Noracle and gas optimizations

davidbrai:
Changes:
1. Noracle: savings settlement prices in storage to serve as an oracle of recent nouns sales via the auction
2. Gas optimizations


packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,112 @@
...
+ * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ *
+ * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ *
+ *********************************/
+
+pragma solidity ^0.8.6;
+
+library Noracle {
+    struct Observation {
+        // The block.timestamp when the auction was settled
+        uint256 blockTimestamp;
```
davidbrai, line 23: let's do some packing here

`blockTimestamp` can be be saved in uint32 (year 2106) or uint34 if we want to be optimistic (year 2514)
`nounId` can be much smaller, uint16? uint32?
`amount` can probably be reduced too, uint80 can represent up to ~1.2M eth if I'm not mistaken

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,112 @@
...
+        // The current highest bid amount
+        uint256 amount;
+        // The address of the auction winner
+        address winner;
+        // whether or not the observation is initialized
+        bool initialized;
+    }
+
+    struct NoracleState {
+        mapping(uint16 => Observation) observations;
```
davidbrai, line 35: any reason to restrict this to uint16? maybe uint32 instead?

packages/nouns-contracts/contracts/libs/Noracle.sol
davidbrai: ideally this could be just one storage slot, but I'm not sure if that works.
we need at least 160 bits for the address, so that leaves 96 bits, which I think is not enough for the rest:

smallest values would be:
timestamp ~ 32 bits
amount (if we don't reduce decimals): ~ 70 (< ~1k ETH)
nounId ~ 16 bits
initialized ~ 1 bit
so at least ~ 119 bits

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,112 @@
...
+        });
+    }
+
+    function grow(NoracleState storage self, uint16 next) internal returns (uint16) {
+        uint16 current = self.cardinalityNext;
+
+        // no-op if the passed next value isn't greater than the current next value
+        if (next <= current) return current;
+
+        // store in each slot to prevent fresh SSTOREs in swaps
```
davidbrai, line 75: swaps? guessing that's copy pasta leftover

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,112 @@
...
+        returns (Observation[] memory observations)
+    {
+        uint16 cardinality = self.cardinality;
+        require(fromAuctionsAgo <= cardinality, 'too many auctions ago');
+
+        observations = new Observation[](fromAuctionsAgo);
+        uint16 initializedObservationsFound = 0;
+        uint16 checkedIndexesCount = 0;
+        while (initializedObservationsFound < fromAuctionsAgo && checkedIndexesCount < cardinality) {
+            checkedIndexesCount++;
```
davidbrai, line 96: assuming the first element of `observations` should be the most recent auction, I think the `checkedIndexesCount++` should move to after the `checkIndex = ...`

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    function write(
+        NoracleState storage self,
+        uint256 blockTimestamp,
+        uint256 nounId,
+        uint256 amount,
+        address winner
+    ) internal {
+        // if the conditions are right, we can bump the cardinality
+        if (self.cardinalityNext > self.cardinality && self.index == (self.cardinality - 1)) {
```
davidbrai, line 55: cache `self.cardinalityNext`

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,112 @@
...
+            self.cardinality = self.cardinalityNext;
+        }
+
+        self.index = (self.index + 1) % self.cardinality;
+        self.observations[self.index] = Observation({
```
davidbrai, line 60: cache `self.index`

packages/nouns-contracts/contracts/libs/Noracle.sol
eladmallel: perhaps a good way to solve this challenge is to give up on plenty of decimals for ETH
instead of 18 decimals we could use say 8 decimals, then we can hold 1K ETH in 40 bits
wdyt?

eladmallel: just pushed some bit packing changes and a test for the amount accuracy, lmk what you think

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,112 @@
...
+            uint16 checkIndex = (self.index + (cardinality - checkedIndexesCount)) % cardinality;
+            Observation storage obs = self.observations[checkIndex];
+            if (obs.initialized) {
+                observations[initializedObservationsFound] = obs;
+                initializedObservationsFound++;
+            }
+        }
+    }
+
+    function warmUpObservation(Observation storage obs) private {
```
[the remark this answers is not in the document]
eladmallel: fixed!

packages/nouns-contracts/contracts/libs/Noracle.sol
davidbrai: I would prefer if we kept the original amount somehow.
let's keep this issue open for now

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -93,20 +93,17 @@ library Noracle {
         uint16 initializedObservationsFound = 0;
         uint16 checkedIndexesCount = 0;
         while (initializedObservationsFound < fromAuctionsAgo && checkedIndexesCount < cardinality) {
-            checkedIndexesCount++;
             uint16 checkIndex = (self.index + (cardinality - checkedIndexesCount)) % cardinality;
             Observation storage obs = self.observations[checkIndex];
             if (obs.initialized) {
                 observations[initializedObservationsFound] = obs;
                 initializedObservationsFound++;
```
davidbrai, line 115: quite minor, but I believe `++i` is cheaper in gas than `i++`

packages/nouns-contracts/contracts/libs/Noracle.sol
eladmallel: yes, perhaps `blockTimestamp > 1` is the easiest condition
could also use nounId, but that's harder to reuse a new nouns fork

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -115,6 +115,14 @@ library Noracle {
...
             }
         }
+
+        if (auctionCount > initializedObservationsFound) {
```
davidbrai, line 119: general feedback: I feel like this function is a bit "fat". wonder if we can trim it.

is the `if (obs.initialized)` actually needed?
in which case is there an uninitialized member? wouldn't we need to break in that case?

packages/nouns-contracts/contracts/libs/Noracle.sol
eladmallel: changed here https://github.com/nounsDAO/nouns-monorepo/pull/598/commits/5468d13a2ceacac4cca2fd7266a5ecfb2fd0c52c

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -103,6 +103,10 @@ library Noracle {
         }
     }
 
+    function ethPriceToUint40(uint256 ethPrice) internal pure returns (uint40) {
+        return uint40(ethPrice / 1e10);
```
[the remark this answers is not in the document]
eladmallel: the latest code with uint48 and 8 decimals holds up to `2814749.76710655` ETH

packages/nouns-contracts/contracts/libs/Noracle.sol
eladmallel: you're absolutely right, my bad.
the case of encountering uninitialized items is when we increase cardinality, and as we're iterating backwards, we loop back to the end where the new cardinality items are all in that state.

eladmallel: did some cleanup here, lmk wdyt: https://github.com/nounsDAO/nouns-monorepo/pull/598/commits/b8557264a97216dfe1bb35a9c8705a0439fe35eb

packages/nouns-contracts/contracts/libs/Noracle.sol
eladmallel: changed as part of this change https://github.com/nounsDAO/nouns-monorepo/pull/598/commits/b8557264a97216dfe1bb35a9c8705a0439fe35eb

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    function write(
+        NoracleState storage self,
+        uint32 blockTimestamp,
+        uint16 nounId,
+        uint48 amount,
+        address winner
+    ) internal {
+        uint32 cardinalityNext = self.cardinalityNext;
+        if (cardinalityNext == 0) revert NotInitialized();
```
davidbrai, line 61: maybe we can return instead of reverting?
the auction house can function without the oracle being initialized

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,140 @@
...
+
+        // if the conditions are right, we can bump the cardinality
+        uint32 currentIndex = self.index;
+        if (cardinalityNext > self.cardinality && currentIndex == (self.cardinality - 1)) {
```
davidbrai, line 65: there's more gas optimizations possible here.

I think you can do a single storage read to read index, cardinality and cardinalityNext, they should be in the same storage slot

davidbrai:
can you add some info about the gas overhead for settling the auction?

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,140 @@
...
+    function grow(NoracleState storage self, uint32 next) internal returns (uint32) {
+        uint32 current = self.cardinalityNext;
+        if (current == 0) revert NotInitialized();
+
+        // no-op if the passed next value isn't greater than the current next value
+        if (next <= current) return current;
+
+        // store in each slot to prevent fresh SSTOREs
+        // this data will not be used because the initialized boolean is still false
+        for (uint32 i = current; i < next; i++) {
```
davidbrai, line 88: gas: `++i`

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,294 @@
...
+        (bool success, ) = to.call{ value: value, gas: 30_000 }(new bytes(0));
+        return success;
+    }
+
+    function initializeOracle() external onlyOwner {
+        oracle.initialize();
+    }
+
+    function growPriceHistory(uint32 newCardinality) external {
+        uint32 current = oracle.cardinalityNext;
```
davidbrai, line 285: gas: `oracle.cardinalityNext` is read again in `oracle.grow`
maybe oracle.grow can return both current and next to avoid the extra sload?

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -0,0 +1,140 @@
...
+            // when cardinality has been increased, and not used up yet, there are uninitialized items
+            // this loop reads from index backwards, so when it hits an uninitialized item
+            // it means all items from there backwards until index are uninitialized, so we should break.
+            if (!initialized(obs)) break;
+
+            observations[observationsCount] = obs;
+            ++observationsCount;
+        }
+
+        if (auctionCount > observationsCount) {
```
davidbrai, line 120: wondering if it's better to drop this block and return some extra zero array members.
or if there's a better way to deal with this.
still thinking about it

packages/nouns-contracts/contracts/libs/Noracle.sol
eladmallel: thanks, fixing

packages/nouns-contracts/contracts/libs/Noracle.sol
eladmallel: yes, I am also not very comfortable with the current state
AFAIK solidity still doesn't support dynamic-size arrays
and we don't have enough information at the beginning to know exactly how many items to allocate

btw, UniV3's observe function is also not very cheap on gas
https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/Oracle.sol#L314

eladmallel:
> can you add some info about the gas overhead for settling the auction?

I have a stashed test that compares v1 and v2, here's the output:
```
v1 gas:  345431
v2 gas:  347869
diff: 2438
```

packages/nouns-contracts/contracts/libs/Noracle.sol
solimander: Should be able to "shorten" by modifying the length:

```solidity
if (auctionCount > observationsCount) {
    assembly {
        mstore(observations, observationsCount)
    }
}
```

eladmallel: great feedback, fixed!

davidbrai: @solimander nice hack :)

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -95,13 +80,7 @@ library Noracle {
     ) internal {
         uint32 currentIndex = self.index;
         uint32 cardinality = self.cardinality;
-        uint32 cardinalityNext = self.cardinalityNext;
-        if (cardinalityNext == 0) return;
-
-        // if the conditions are right, we can bump the cardinality
-        if (cardinalityNext > cardinality && currentIndex == (cardinality - 1)) {
```
davidbrai: @eladmallel are you sure `cardinalityNext` is not needed?
My understanding is that it's needed so that `observe` works correctly.

what happens in this case:
`self.cardinality = 10`
and `self.index = 0`
assuming all the 10 slots are filled with data

then we call `grow(20)`
now slows 10-19 are empty
but if we call `observe` it will look at slot 0, then slot 19 (going backwards), see an empty slot, and quit.
that's not what we'd want, right?

packages/nouns-contracts/contracts/libs/Noracle.sol
```diff
@@ -98,11 +102,13 @@ library Noracle {
         uint32 cardinalityNext = self.cardinalityNext;
         if (cardinalityNext == 0) return;
 
-        // if the conditions are right, we can bump the cardinality
+        // if the conditions are right, we can bump the cardinality
```
davidbrai, line 105: nit: extra whitespace

packages/nouns-contracts/contracts/NounsAuctionHousePreV2Migration.sol
```diff
@@ -0,0 +1,52 @@
...
+    uint256 public timeBuffer;
+    uint256 public reservePrice;
+    uint8 public minBidIncrementPercentage;
+    uint256 public duration;
+    INounsAuctionHouse.Auction public auction;
+
+    function migrate() public onlyOwner {
+        INounsAuctionHouse.Auction memory _auction = auction;
+
+        auction = INounsAuctionHouse.Auction(0, 0, 0, 0, payable(address(0)), false);
```
davidbrai, line 38: minor: add inline comment that this for nullifying the previous storage location

packages/nouns-contracts/contracts/NounsAuctionHousePreV2Migration.sol
```diff
@@ -0,0 +1,52 @@
...
+
+        INounsAuctionHouse.AuctionV2 storage auctionV2;
+        assembly {
+            auctionV2.slot := auction.slot
```
davidbrai, line 42: @eladmallel any particular reason to reuse the same slot?
alternatively we could keep the old `auction` slot and mark it as deprecated.
so I see the reason could be to keep the V2 contract cleaner. is that the reasoning or is there something else?

davidbrai:
@eladmallel perhaps we want to consider the public API of the `auction` state?
specifically I can think of the autobidder contract as one potential that can break. don't know if there are others.
we should take a look, and also consider create a view function that is not dependent on the storage layout

packages/nouns-contracts/test/foundry/NounsAuctionHouseV2.t.sol
```diff
@@ -0,0 +1,382 @@
...
+        vm.deal(bidder, bid);
+        vm.prank(bidder);
+        auction.createBid{ value: bid }(nounId);
+        vm.warp(endTime);
+        auction.settleCurrentAndCreateNewAuction();
+        return block.timestamp;
+    }
+}
+
+contract NounsAuctionHouseV2_OracleTest is NounsAuctionHouseV2Test {
```
davidbrai, line 237: @eladmallel by inheriting from `NounsAuctionHouseV2Test` it's also running all the tests again.
better to inherit from a contract without test methods

packages/nouns-contracts/contracts/NounsAuctionHousePreV2Migration.sol
eladmallel: definitely the reason is to keep V2 cleaner

packages/nouns-contracts/test/foundry/NounsAuctionHouseV2.t.sol
eladmallel: good catch, fixing now

solimander:
There are several additional gas optimizations that could be made to reduce settlement by 6K+ gas and bid creation by 4K+ gas:

- Convert `nouns`, `weth`, and `duration` to immutable variables. Currently, they can only be upgraded via a new logic contract anyways.
- Collapse `timeBuffer`, `reservePrice`, and `minBidIncrementPercentage` into a single slot
- Pack `paused` into one of the slots with remaining space that are already read during settlement.

It may be too complex to migrate the storage for these, but I'm curious on your take.

Possible to add `uint256[5] private __gap;` where the existing storage values are and pack below.

PR #600, opened 2022-10-21
Enable Refundable Voting in the Frontend

solimander:


packages/nouns-webapp/src/wrappers/nounsDao.ts
```diff
@@ -493,19 +494,42 @@ export const useCastVoteWithReason = () => {
...
     'castRefundableVote',
   );
-  return { castRefundableVote, castRefundableVoteState };
+
+  return {
+    castRefundableVote: async (...args: any[]): Promise<void> => {
+      const contract = connectContractToSigner(nounsDaoContract, undefined, library);
+      const gasLimit = await contract.estimateGas.castRefundableVote(...args);
+      return castRefundableVote(...args, {
+        gasLimit: gasLimit.add(20_000), // A 20,000 gas pad is used to avoid 'Out of gas' errors
```
davidbrai, line 508: @solimander why is this needed? why isn't `estimateGas` returning the correct number?

solimander: Not sure of the exact cause in this case. Usually this occurs because the blockchain state has changed between the estimate and execution, but that should not be a factor here. We had to pad the gas estimate for `createBid` to prevent "Out of Gas" errors as well. Both do contain an ETH transfer at the end of the transaction, so that may play some role.

Regardless, an additional buffer is needed to prevent random "Out of Gas" errors; even when voting directly from an EOA.

solimander: Unfortunately, Gnosis Safes don't seem to use the gas limit that's passed to them, so they still underestimate. Need to see what our options are there.

PR #630, opened 2022-11-22
fix typo & icon change on mobile nav

ripe0x:
- the playground link in the mobile nav was linking to `/explore` instead of `/playground`. this has been resolved. 
- the mobile  `Explore` button now uses the ⌐◨-◨ icon instead of the font awesome star.


lastpunk9999 (approved):
Good catch. Thanks!

PR #643, opened 2022-12-10
docs(webapp): add build command to readme

DanielDanielsson:
Building all packages is requierd in order to run the webapp


solimander:
Thanks for the contribution, but `yarn` _should_ trigger the `prepare` hook, which builds all necessary packages. Is that not the case on your machine?

PR #646, opened 2022-12-16
Markdown Table Support

solimander:


volkyeth:
Will be taken care of in #1070 
Also need to ensure feature parity (and some standardization) with camp

PR #651, opened 2022-12-22
Brianj add delegation warning and streaming payment type

bcjgit:
# TLDR

This PR does 3 things:

1. Adds warning to delegation flow to warn users if their delegation would decrease their vote power enough to the point where active props would be cancelable
2. Adds a new streaming payment type to the proposal creation flow (supporting WETH + USDC). This includes both code for the UI flow and the code to construct the resulting transactions (not just creating the stream, but also converting ETH to WETH, getting USDC from token buyer, etc.)
3. Adds UI to the prop page (`/vote/:id`) such that when the recipient of a stream goes to that page after a prop has been executed they will see a modal which will allow them to claim the ETH they are owed

# Testing done

Created USDC and ETH streams on testnet and successfully claimed them

# Open Questions 

1. Maybe this betrays my lack of understanding of the inner workings of the contracts here ... but not sure if there's anything else I need to do on my end to make sure that USDC gets to the streams even if there isn't any USDC in the token buyer at the time of execution


packages/nouns-webapp/src/config.ts
```diff
@@ -90,30 +93,42 @@ const externalAddresses: Record<SupportedChains, ExternalContractAddresses> = {
...
+    nounsStreamFactory: undefined,
   },
   [ChainId.Goerli]: {
     lidoToken: '0x2DD6530F136D2B56330792D46aF959D9EA62E276',
     usdcToken: '0x07865c6e87b9f70255377e024ace6630c1eaa37f',
+    weth: '0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6',
     payerContract: '0xD4A3bf1dF54699E63A2ef7F490E8E22b27B945f0',
     tokenBuyer: '0x61Ec4584c5B5eBaaD9f21Aac491fBB5B2ff30779',
     chainlinkEthUsdc: undefined,
+    nounsStream: '0xb82c6b4889969deba04d2ade4687a4656186895d',
```
davidbrai, line 110: @bcjgit I think this is unused, can be removed. the frontend should only need to interact with the StreamFactory or new stream clones

packages/nouns-webapp/src/pages/Vote/index.tsx
```diff
@@ -340,6 +356,45 @@ const VotePage = ({
...
+            .map(txn => {
+              const parsedCallData = parseStreamCreationCallData(txn.callData);
+              if (parsedCallData.recipient.toLowerCase() !== account?.toLowerCase()) {
+                return <></>;
+              }
+
+              return (
+                <Row className={clsx(classes.section, classes.transitionStateButtonSection)}>
+                  <span className={classes.boldedLabel}>
+                    <Trans>Only visable to you</Trans>
```
davidbrai, line 371: type: visable -> visible

packages/nouns-webapp/src/locales/en-US.po
```diff
@@ -202,9 +220,14 @@ msgstr "Auction ended"
 msgid "Auction ends in"
 msgstr "Auction ends in"
 
+#: src/components/StreamWithdrawModal/index.tsx
+msgid "Avilable to withdraw"
+msgstr "Avilable to withdraw"
```
davidbrai, line 225: type in both: Avilable -> Available

packages/nouns-webapp/src/components/StreamWithdrawModal/index.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { SupportedCurrency } from '../ProposalActionsModal/steps/TransferFundsDetailsStep';
+import ModalLabel from '../ModalLabel';
+
+dayjs.extend(relativeTime);
+
+export const Backdrop: React.FC<{ onDismiss: () => void }> = props => {
+  return <div className={classes.backdrop} onClick={props.onDismiss} />;
+};
+
+const StreamWidthdrawModalOverlay: React.FC<{
```
davidbrai, line 32: type in many places:
StreamWidthdrawModal -> StreamWithdrawModal

solimander (approved):
🚀

PR #674, opened 2023-02-06
Display Raw Calldata in Proposal Actions

solimander:
Display raw calldata in proposal calls that have data, but no signature.


iainnash:
Looks good to me!

The other issue you can run into is a bad proposal that uses the wrong signature for the transaction.

Maybe there can be a validation step that hashes the signature and matches the txn id with the passed in txn id and falls back to this view if it fails?

solimander:
Great idea, @iainnash. Pushing this PR up as a quick fix, but I'll create an issue for that addition.

PR #677, opened 2023-02-09
DRAFT: dao v3 contracts

davidbrai:


packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
```diff
@@ -0,0 +1,685 @@
...
+            address signer = signers[i] = proposerSignatures[i].signer;
+
+            checkNoActiveProp(ds, signer);
+            ds.latestProposalIds[signer] = proposalId;
+            votes += ds.nouns.getPriorVotes(signer, block.number - 1);
+        }
+
+        checkNoActiveProp(ds, msg.sender);
+        ds.latestProposalIds[msg.sender] = proposalId;
+        votes += ds.nouns.getPriorVotes(msg.sender, block.number - 1);
```
davidbrai, line 171: @eladmallel `proposal.proposer` votes should also be counted `cancel`

packages/nouns-subgraph/src/custom-types/ParsedProposalV3.ts
```diff
@@ -0,0 +1,102 @@
...
+  quorumVotes: BigInt = BIGINT_ZERO;
+  description: string = '';
+  title: string = '';
+  status: string = '';
+  signers: string[] = [];
+
+  static fromV1Event(event: ProposalCreatedWithRequirements1): ParsedProposalV3 {
+    const proposal = new ParsedProposalV3();
+
+    proposal.id = event.params.id.toString();
```
davidbrai, line 31: @eladmallel consider extracting all the parts identical to the function below

packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -29,63 +32,69 @@ import {
...
-export function handleProposalCreatedWithRequirements(
+export function handleProposalCreatedWithRequirementsV3(
   event: ProposalCreatedWithRequirements,
 ): void {
-  let proposal = getOrCreateProposal(event.params.id.toString());
-  let proposer = getOrCreateDelegateWithNullOption(event.params.proposer.toHexString(), false);
+  handleParsedProposalV3(ParsedProposalV3.fromV3Event(event));
+}
+
+function handleParsedProposalV3(parsedProposal: ParsedProposalV3): void {
```
davidbrai, line 50: @eladmallel maybe rename to something like `handleProposerCreated` ? `handleParsedProposalV3` is not very clear

packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -29,63 +32,69 @@ import {
...
+  proposal.description = parsedProposal.description;
+  proposal.title = parsedProposal.title;
+  proposal.status = parsedProposal.status;
+
+  const signerDelegates = new Array<string>(parsedProposal.signers.length);
+  for (let i = 0; i < parsedProposal.signers.length; i++) {
+    const signerAddress = parsedProposal.signers[i];
+    const signerDelegate = getOrCreateDelegateWithNullOption(signerAddress, false);
+
+    if (signerDelegate == null) {
```
davidbrai, line 90: @eladmallel in which case would `signerDelegate` be `null`? shouldn't it always be not null because it's returned from `getOrCreate`?

packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -100,6 +109,35 @@ export function handleProposalCreatedWithRequirements(
   proposal.save();
 }
 
+export function handleProposalUpdated(event: ProposalUpdated): void {
+  const updateId = event.params.id
+    .toString()
+    .concat('-')
+    .concat(event.transaction.hash.toHexString());
```
davidbrai, line 118: @eladmallel unlikely to have collisions, but safer to also concat `event.logIndex.toString()`

packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -100,6 +109,35 @@ export function handleProposalCreatedWithRequirements(
...
+
+  const proposalUpdate = getOrCreateProposalUpdate(updateId);
+  const proposal = getOrCreateProposal(event.params.id.toString());
+
+  // First save the current state of the proposal to the updates history entity
+  proposalUpdate.createdAt = event.block.timestamp;
```
davidbrai, line 124: @eladmallel I think `ProposalUpdate` should point to `Proposal`, i.e have a field:
`proposal: Proposal!`

packages/nouns-contracts/contracts/governance/NounsDAOV3Votes.sol
```diff
@@ -0,0 +1,354 @@
...
+            uint256 basefee = min(block.basefee, MAX_REFUND_BASE_FEE);
+            uint256 gasPrice = min(tx.gasprice, basefee + MAX_REFUND_PRIORITY_FEE);
+            uint256 gasUsed = min(startGas - gasleft() + REFUND_BASE_GAS, MAX_REFUND_GAS_USED);
+            uint256 refundAmount = min(gasPrice * gasUsed, balance);
+            (bool refundSent, ) = msg.sender.call{ value: refundAmount }('');
+            emit RefundableVote(msg.sender, refundAmount, refundSent);
+        }
+    }
+
+    function proposalCreationBlock(NounsDAOStorageV3.StorageV3 storage ds, NounsDAOStorageV3.Proposal storage proposal)
```
davidbrai, line 329: @eladmallel maybe rename to `proposalVoteSnapshotBlock` ?

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
eladmallel: fixed!

packages/nouns-subgraph/src/custom-types/ParsedProposalV3.ts
eladmallel: @davidbrai the main issue is that these two event types are auto generated without a shared super type with all the common fields, that's why I wrote out these two separate functions. also note that the way event params are handled is by using an array inside where each field is a certain index, and they don't overlap in the autogenerated code, making it even harder to refactor and reuse, so I just opted for code duplication.

wdyt?

packages/nouns-subgraph/src/nouns-dao.ts
eladmallel: fixed!

packages/nouns-subgraph/src/nouns-dao.ts
eladmallel: actually it might happen, since we don't really enforce that every single signer on a prop have at least one vote.. we actually let signers with zero votes take part.

btw, we also allow people with zero votes to cast a vote :)

key question:
should we enforce that all signers on a proposal have at least one vote?

if we say that we should enforce:
should we enforce the same thing when casting votes? and if not, what's the difference?

if we aren't enforcing:
should subgraph create a Delegate entity for them as we do in the propose and vote handlers? 
or should we leave the code as-is and not create a new delegate for such signers?

wdyt @davidbrai?

eladmallel: in the meantime I'm updating the code to create a zero votes delegate for any signer without a delegate entity, to at least be consistent with how we do things in other places here. still important to discuss these questions together!

packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -100,6 +109,35 @@ export function handleProposalCreatedWithRequirements(
...
+  proposalUpdate.description = proposal.description;
+  proposalUpdate.save();
+
+  // Then update the proposal to the latest state
+  proposal.targets = changetype<Bytes[]>(event.params.targets);
+  proposal.values = event.params.values;
+  proposal.signatures = event.params.signatures;
+  proposal.calldatas = event.params.calldatas;
+  proposal.description = event.params.description.split('\\n').join('\n');
+  proposal.title = extractTitle(proposal.description);
```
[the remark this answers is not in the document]
eladmallel: great, added!

packages/nouns-subgraph/src/nouns-dao.ts
eladmallel: added

packages/nouns-subgraph/src/nouns-dao.ts
eladmallel: of course, total oopsie on my end :) fixed!

packages/nouns-contracts/contracts/governance/NounsDAOV3Votes.sol
eladmallel: fixed

packages/nouns-subgraph/src/custom-types/ParsedProposalV3.ts
davidbrai: @eladmallel yeah I see the issue. I guess assemblyscript is pretty rigid on the types

packages/nouns-subgraph/src/nouns-dao.ts
davidbrai: @eladmallel what I meant is that the call your making to `getOrCreateDelegateWithNullOption` seems to always return a non-null object:
```
if (delegate == null && createIfNotFound) {
    delegate = new Delegate(id);
    ...
    return delegate;
}
```

so if I'm not mistaken, the `if (signerDelegate == null) {` check is never true

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
```diff
@@ -537,10 +543,11 @@ contract NounsDAOStorageV3 {
         /// @notice Pending new vetoer
         address pendingVetoer;
         // ================ V3 ================ //
-        uint256 lastMinuteWindowInBlocks;
-        uint256 objectionPeriodDurationInBlocks;
         /// @notice user => sig => isCancelled: signatures that have been cancelled by the signer and are no longer valid
         mapping(address => mapping(bytes32 => bool)) cancelledSigs;
+        uint32 lastMinuteWindowInBlocks;
+        uint32 objectionPeriodDurationInBlocks;
+        uint32 proposalUpdatablePeriodInBlock;
```
davidbrai, line 550: @eladmallel block -> blocks

packages/nouns-subgraph/src/nouns-dao.ts
eladmallel: @davidbrai I heard you, and tried making it better in this commit: https://github.com/nounsDAO/nouns-monorepo/pull/677/commits/984a3cc1cb81eb15b6363c60dfb2753230b62cb7

wdyt?

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
eladmallel: fixed!

packages/nouns-subgraph/src/nouns-dao.ts
davidbrai: lgtm

packages/nouns-contracts/contracts/governance/NounsDAOProxyV3.sol
```diff
@@ -0,0 +1,156 @@
...
+// Proxy.sol source code licensed under MIT License.
+//
+// MODIFICATIONS
+// The fallback() and receive() functions of Proxy.sol have been used to allow Solidity > 0.6.0 compatibility
+
+pragma solidity ^0.8.6;
+
+import './NounsDAOInterfaces.sol';
+
+contract NounsDAOProxyV3 is NounsDAOProxyStorage, NounsDAOEvents {
```
davidbrai, line 43: need to update `deploy-local-dao-v3.ts` to use proxyv3 instead of proxyv2

packages/nouns-contracts/contracts/governance/NounsDAOProxyV3.sol
```diff
@@ -0,0 +1,156 @@
...
+    struct ProxyParams {
```
davidbrai, line 44: @eladmallel what's the purpose of the struct?

packages/nouns-contracts/test/foundry/NounsDAOLogicV3.t.sol
```diff
@@ -0,0 +1,383 @@
...
+        // Active
+        vm.roll(block.number + VOTING_DELAY);
+        assertTrue(dao.state(proposalId) == NounsDAOStorageV3.ProposalState.Active);
+        vm.expectRevert(abi.encodeWithSelector(NounsDAOV3Proposals.CanOnlyEditUpdatableProposals.selector));
+        updateProposal(proposer, proposalId, makeAddr('target'), 0, '', '', '');
+
+        // Succeeded
+        vm.prank(proposer);
+        dao.castVote(proposalId, 1);
+        vm.roll(block.number + VOTING_DELAY + VOTING_PERIOD);
```
davidbrai, line 259: @eladmallel I think here you need to roll only `block.number + VOTING_PERIOD` because you already rolled `VOTING_DELAY` blocks before

packages/nouns-contracts/contracts/governance/NounsDAOProxyV3.sol
eladmallel: @davidbrai I am open to not having it. again here my main motivation is to have a proxy that's ready for anyone else to deploy V3 directly and felt right for them to be able to set all the config variables in one function call.

no strong opinion here, can go with whatever you prefer, lmk!

packages/nouns-contracts/contracts/governance/NounsDAOProxyV3.sol
eladmallel: @davidbrai to avoid stack too deep in the proxy constructor :)

packages/nouns-contracts/test/foundry/NounsDAOLogicV3.t.sol
eladmallel: removed!

packages/nouns-contracts/test/foundry/NounsDAOLogicV3/NounsDAOLogicV3BaseTest.sol
```diff
@@ -153,17 +153,59 @@ abstract contract NounsDAOLogicV3BaseTest is Test, DeployUtils, SigUtils {
         string memory description,
```
davidbrai, line 167: @eladmallel maybe this file should also be `.t.sol`?

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
```diff
@@ -0,0 +1,721 @@
...
+        view
+        returns (uint256 propThreshold)
+    {
+        uint256 totalSupply = ds.nouns.totalSupply();
+        propThreshold = bps2Uint(ds.proposalThresholdBPS, totalSupply);
+
+        require(votes > propThreshold, 'NounsDAO::propose: proposer votes below proposal threshold');
+    }
+
+    function checkProposaTxs(ProposalTxs memory txs) internal pure {
```
davidbrai, line 683: @eladmallel typo: `checkProposa*l*Txs`

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -0,0 +1,272 @@
...
+     *   STATE
+     * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
+     */
+
+    /// @notice The cost non-Nouners must pay in ETH in order to emit a new proposal candidate event from this contract.
+    uint256 public createCandidateCost;
+    /// @notice The cost non-Nouners must pay in ETH in order to emit a proposal candidate update event from this contract.
+    uint256 public updateCandidateCost;
+    /// @notice The Nouns token contract.
+    NounsTokenLike public nounsToken;
```
davidbrai, line 88: @eladmallel this can probably be immutable?
would need to be set in constructor instead of initializer then

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -0,0 +1,272 @@
...
+     * @param encodedPropHash the keccak256 hash of the candidate version signed; the hash is performed on the output of
+     * the `NounsDAOV3Proposals.calcProposalEncodeData` function.
+     */
+    function addSignature(
+        address signer,
+        bytes memory sig,
+        uint256 expirationTimestamp,
+        bytes32 encodedPropHash,
+        string memory reason
+    ) external {
```
davidbrai, line 226: @eladmallel
1. do we want to add signature validation here?
2. should this also specify the slug?

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -0,0 +1,272 @@
...
+        bytes[] memory calldatas,
+        string memory description,
+        string memory slug
+    ) external payable {
+        if (!isNouner(msg.sender) && msg.value < createCandidateCost) revert MustBeNounerOrPaySufficientFee();
+        if (propCandidates[msg.sender][keccak256(bytes(slug))]) revert SlugAlreadyUsed();
+
+        propCandidates[msg.sender][keccak256(bytes(slug))] = true;
+
+        bytes memory encodedProp = NounsDAOV3Proposals.calcProposalEncodeData(
```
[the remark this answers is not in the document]
eladmallel: changed to internal to embed it

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
eladmallel: agreed, will do

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
eladmallel: as discussed:

1. leaving out for now, to save on gas 
2. the slug is part of the function and the event it emits 
3. will do to minimize spam

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -0,0 +1,272 @@
...
+     * @param reason their free text feedback.
+     */
+    function sendFeedback(
+        uint256 proposalId,
+        uint8 support,
+        string memory reason
+    ) external {
+        if (!isNouner(msg.sender)) revert MustBeNouner();
+
+        emit FeedbackSent(msg.sender, proposalId, support, reason);
```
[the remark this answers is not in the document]
eladmallel: sounds good, will refactor to add weight to the event

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
eladmallel: done

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
eladmallel: ended up doing signature validation because it's the only way of making sure the signer is indeed the account submitting the signature.

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -266,6 +269,22 @@ contract NounsDAOData is OwnableUpgradeable {
         emit UpdateCandidateCostSet(oldUpdateCandidateCost, newUpdateCandidateCost);
     }
 
+    /**
+     * @notice Withdraw ETH from this contract's balance. Only owner can call this function.
+     * @param to the recipient.
+     * @param amount the ETH amount to send.
+     */
+    function withdrawETH(address to, uint256 amount) external onlyOwner {
+        if (amount > address(this).balance) revert AmountExceedsBalance();
```
davidbrai, line 310: @eladmallel I think we can drop this. it will revert anyway

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -212,24 +214,44 @@ contract NounsDAOData is OwnableUpgradeable {
...
 
-        emit SignatureAdded(msg.sender, signer, sig, expirationTimestamp, proposer, slug, encodedPropHash, reason);
+        bytes32 sigDigest = NounsDAOV3Proposals.sigDigest(
+            NounsDAOV3Proposals.PROPOSAL_TYPEHASH,
+            encodedProp,
+            expirationTimestamp,
+            address(this)
+        );
+
+        if (!SignatureChecker.isValidSignatureNow(msg.sender, sigDigest, sig)) revert InvalidSignature();
```
davidbrai, line 251: (adding this as a reminder)
in the context of signing from a gnosis safe:
the safe operator will need to wait until the signature tx is included in a block, and only then they can call this function.
we may want to give some clarity around this in the UI

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
```diff
@@ -256,6 +258,70 @@ library NounsDAOV3Proposals {
...
+        emit ProposalDescriptionUpdated(proposalId, msg.sender, description, updateMessage);
+    }
+
+    function checkProposalUpdatable(
+        NounsDAOStorageV3.StorageV3 storage ds,
+        uint256 proposalId,
+        NounsDAOStorageV3.Proposal storage proposal
+    ) internal view {
+        // TODO: gas: does reading the proposal once save gas?
+        if (state(ds, proposalId) != NounsDAOStorageV3.ProposalState.Updatable) revert CanOnlyEditUpdatableProposals();
```
eladmallel, line 321: @davidbrai maybe we should refactor state such that there's a version of it that takes a `Proposal storage proposal` parameter so we avoid the double reading we perform when we're holding a proposal in a function and then re-reading it when calling state?

eladmallel: I see the related todo you added as a comment above :) I think it's probably not very hard for you to look into it as you're writing some unit tests for these functions?

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
```diff
@@ -160,7 +172,7 @@ library NounsDAOV3Proposals {
         string memory description
     ) external returns (uint256) {
         if (proposerSignatures.length == 0) revert MustProvideSignatures();
-        checkProposaTxs(txs);
+        checkProposalTxs(txs);
```
eladmallel, line 175: @davidbrai thank you for the rename!

packages/nouns-contracts/contracts/governance/NounsDAOLogicV3.sol
```diff
@@ -223,6 +223,31 @@ contract NounsDAOLogicV3 is NounsDAOStorageV3, NounsDAOEventsV3 {
...
+     * @param proposalId proposal's id
+     * @param description the updated description
+     * @param updateMessage short message to explain the update
+     */
+    function updateProposalDescription(uint256 proposalId, string calldata description, string calldata updateMessage) external {
+        ds.updateProposalDescription(proposalId, description, updateMessage);
+    }
+
+    /**
+     * Updates the proposal's transactions. Only the proposer can update it, and only during the updateable period.
```
eladmallel, line 237: @davidbrai intentionally not verbose on params natspecs?

packages/nouns-subgraph/src/nouns-dao-data.ts
```diff
@@ -0,0 +1,161 @@
...
+  const candidate = getOrCreateProposalCandidate(candidateId);
+
+  candidate.targets = changetype<Bytes[]>(event.params.targets);
+  candidate.values = event.params.values;
+  candidate.signatures = event.params.signatures;
+  candidate.calldatas = event.params.calldatas;
+  candidate.description = event.params.description.split('\\n').join('\n');
+  candidate.title = extractTitle(candidate.description);
+  candidate.encodedProposalHash = event.params.encodedProposalHash;
+  candidate.createdTransactionHash = event.transaction.hash;
```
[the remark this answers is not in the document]
eladmallel: nope, good catch, removing

packages/nouns-contracts/contracts/governance/NounsDAOLogicV3.sol
davidbrai: yeah, felt self explanatory

PR #688, opened 2023-02-27
add korean language support

fiveoceans-dev:
files modified:
1. [locales.ts](https://github.com/fiveoceans-dev/nouns-monorepo/blob/master/packages/nouns-webapp/src/i18n/locales.ts)
2. [NounsI18nProvider.tsx](https://github.com/fiveoceans-dev/nouns-monorepo/blob/master/packages/nouns-webapp/src/i18n/NounsI18nProvider.tsx)

files added:
1. [ko-KOR.po](https://github.com/fiveoceans-dev/nouns-monorepo/blob/master/packages/nouns-webapp/src/locales/ko-KOR.po)

### DIFF
--- a/packages/nouns-webapp/lingui.config.ts
+++ b/packages/nouns-webapp/lingui.config.ts
@@ -17,7 +17,7 @@ const linguiConfig = {
   formatOptions: {
     lineNumbers: false,
   },
-  locales: ['en-US', 'ja-JP', 'pseudo'],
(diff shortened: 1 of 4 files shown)


fiveoceans-dev:
@solimander
Hi, I wonder if it is okay to add korean language support to the nouns.wtf homepage... 
I am okay to cancel PR if it is something that creates more friction than good
Thanks

PR #699, opened 2023-03-30
Explorer

solimander:


volkyeth:
@solimander @gitblamecarrot this is so cool :D
was there a reference design?

solimander:
@volkyeth  hah, totally forgot about this! we were going off a couple initial designs from 4156.

PR #711, opened 2023-06-02
audit: bugfix: proposeBySigs allows multiple signatures from same signer

davidbrai:
this is because when checking if a proposal is already active, the check fails because the proposal hasn't been initialized yet.

this fix initializes the proposal first


packages/nouns-contracts/test/foundry/NounsDAOLogicV3/ProposeBySigs.t.sol
```diff
@@ -440,6 +440,34 @@ contract ProposeBySigsTest is NounsDAOLogicV3BaseTest {
...
+        NounsDAOStorageV3.ProposerSignature[] memory proposerSignatures = new NounsDAOStorageV3.ProposerSignature[](2);
+        proposerSignatures[0] = NounsDAOStorageV3.ProposerSignature(
+            signProposal(proposerWithNoVotes, signerWithVote1PK, txs, 'description', expirationTimestamp, address(dao)),
+            signerWithVote1,
+            expirationTimestamp
+        );
+        proposerSignatures[1] = NounsDAOStorageV3.ProposerSignature(
+            signProposal(proposerWithNoVotes, signerWithVote1PK, txs, 'description', expirationTimestamp, address(dao)),
+            signerWithVote1,
+            expirationTimestamp
```
eladmallel, line 463: seems perhaps important to use a different timestamp, maybe as a separate test?
in this test I believe both sigs will have the same bytes, but the actual test is that it fails given the same signer.

wdyt?

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
```diff
@@ -225,23 +225,27 @@ library NounsDAOV3Proposals {
...
         );
+
+        (uint256 votes, address[] memory signers) = verifySignersCanBackThisProposalAndCountTheirVotes(
+            ds,
+            proposerSignatures,
+            txs,
+            description,
+            proposalId
+        );
+        require(votes > propThreshold, 'NounsDAO::propose: proposer votes below proposal threshold');
```
eladmallel, line 247: since we're no longer reusing the prop threshold check code, maybe better this becomes a custom error, to get less auditing noise in future?

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
```diff
@@ -225,23 +225,27 @@ library NounsDAOV3Proposals {
...
+
         NounsDAOStorageV3.Proposal storage newProposal = createNewProposal(
             ds,
             proposalId,
             propThreshold,
             adjustedTotalSupply,
             txs
         );
+
+        (uint256 votes, address[] memory signers) = verifySignersCanBackThisProposalAndCountTheirVotes(
```
eladmallel, line 240: wonder if it's worth adding an inline comment above this emphasizing this must happen after proposal creation?

packages/nouns-contracts/test/foundry/NounsDAOLogicV3/ProposeBySigs.t.sol
davidbrai: sure, I changed the timestamp

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
davidbrai: sure. in this case I'm changing the error in both places

davidbrai:
@eladmallel please review again

eladmallel:
@davidbrai LGTM

PR #713, opened 2023-06-14
audit: proposeBySigs: exclude signers without voting power

davidbrai:
since they don't contribute voting power, they shouldn't be given the power to cancel

mitigation for: https://github.com/spearbit-audits/review-nouns/issues/17


eladmallel:
LGTM!

PR #714, opened 2023-06-14
audit fix: fork auction griefing

eladmallel:
Fixes this spearbit audit issue: https://github.com/spearbit-audits/review-nouns/issues/9

TLDR of the bug:

1. execute a fork
2. buy new nouns on the original auction
3. join the fork with those new nouns
4. attempt to unpause the fork auction house

the bug is such that the fork auction house can't unpause, because minting fails, because we're trying to mint an ID that was already minted through fork joining.

the fix is when nouns join a fork, to make sure the next ID the fork auction house will use is always higher than highest noun ID to ever join.


davidbrai:
LGTM

PR #719, opened 2023-06-19
fix timestamp check in claimDuringForkPeriod

davidbrai:
now it matches the check in NounsDAOV3Fork.isForkPeriodActive

otherwise it allows the OG DAO to claim in the last second

fix for: https://github.com/spearbit-audits/review-nouns/issues/40


eladmallel:
LGTM

PR #734, opened 2023-06-22
move token-buyer ownership transfer to cleanup proposal

davidbrai:
fix for: https://github.com/spearbit-audits/review-nouns/issues/54

the assumption is that this followup proposal will be created shortly after the main upgrade proposal executes.
this will allow any existing USDC proposals to execute via timelockV1. new USDC proposals create after the upgrade proposal execution will be queued on timelock v2.
the cleanup proposal will change the ownership of token buyer to timelock v2 so new proposals should also be able to execute correctly


eladmallel:
LGTM

PR #735, opened 2023-06-22
split withdrawDAONounsFromEscrow to 2 functions

davidbrai:
fixes: https://github.com/spearbit-audits/review-nouns/issues/61

this is in order to make it clearer that when withdrawing nouns to an address other than the treasury it will increase the total supply, which then affects quorum, proposal threshold, and fork calculations


eladmallel:
LGTM

PR #747, opened 2023-07-05
subgraph: refactor to be able to match submitted proposals with candidates

davidbrai:
extracted a ProposalCandidateContent which has the proposal content hash as its id
This way it can be fetched when a proposal is created


eladmallel:
LGTM @davidbrai

PR #748, opened 2023-07-06
Nouns data contract: support candidates for updateProposeBySigs

davidbrai:
added proposalIdToUpdate, which is zero if the proposal is new


eladmallel:
LGTM @davidbrai

PR #749, opened 2023-07-07
add feedback on candidates

eladmallel:


packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -257,10 +256,28 @@ contract NounsDAOData is OwnableUpgradeable, UUPSUpgradeable, NounsDAODataEvents
         string memory reason
     ) external {
         if (support > 2) revert InvalidSupportValue();
-        uint96 votes = nounsToken.getPriorVotes(msg.sender, block.number - PRIOR_VOTES_BLOCKS_AGO);
```
davidbrai, line 260: @eladmallel thought it could be useful to still include `votes` in the event. Similar to the VoteCast event
I guess clients can get this data by indexing delegation. is that what you had in mind?

davidbrai: I see that's what you implemented. guess it's ok

davidbrai:
LGMT

PR #750, opened 2023-07-07
data: add configurable fee recipient

eladmallel:


packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -174,6 +185,10 @@ contract NounsDAOData is OwnableUpgradeable, UUPSUpgradeable, NounsDAODataEvents
             encodedProp = abi.encodePacked(proposalIdToUpdate, encodedProp);
         }
 
+        if (msg.value > 0) {
+            feeRecipient.sendValue(msg.value);
```
davidbrai, line 189: nit: I think OZ doesn't add anything in this case. for me it just forces me to go into their implementation to see what it does.

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -174,6 +185,10 @@ contract NounsDAOData is OwnableUpgradeable, UUPSUpgradeable, NounsDAODataEvents
             encodedProp = abi.encodePacked(proposalIdToUpdate, encodedProp);
         }
 
+        if (msg.value > 0) {
```
davidbrai, line 188: suggestion: check that `feeRecipient` is not zero?
if `feeRecipient` is zero, then this contract can accumulate eth and the owner can withdraw later

PR #755, opened 2023-07-12
BidWithComment upgrades from Lil Nouns DAO

EricDevito:


EricDevito:
Upgrade to mainnet subgraph required before merge.
Also, testing here (sepolia): https://master--stellular-cat-047478.netlify.app

PR #760, opened 2023-07-17
lock initliazers on upgradable implementation contracts

davidbrai:
following the recommendation mentioned here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.4.0/contracts/proxy/utils/Initializable.sol#L18-L31


eladmallel:
@davidbrai I think it's nice to add a short test that tries to initialize all these implementation contracts and shows they all revert. wdyt?

davidbrai:
@eladmallel added tests. ok to merge?

eladmallel:
@davidbrai LGTM!

PR #762, opened 2023-07-18
bugfix: don't allow quit() in forked dao to pass duplicate erc20 tokens

davidbrai:
fix for: https://github.com/code-423n4/2023-07-nounsdao-findings/issues/180

the bug allowed for the same token to be sent multiple times to the caller


solimander:
LGTM

PR #796, opened 2023-09-28
repo cleanup

davidbrai:
* moving old contracts to archive folder
* deleting old tests


volkyeth:
those contracts seem to have been removed already

PR #799, opened 2023-09-28
AuctionHouseV2: price feed, gas optimizations

eladmallel:


packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /**
+     * @notice Create a bid for a Noun, with a given amount.
+     * @dev This contract only accepts payment in ETH.
+     */
+    function createBid(uint256 nounId) external payable override {
+        INounsAuctionHouseV2.AuctionV2 memory _auction = auction;
+
+        if (_auction.nounId != nounId) revert NounNotUpForAuction();
```
davidbrai, line 126: @eladmallel I've recently started preferring string errors and `require` instead of custom errors. often it's easier to debug with string errors.
what do we gain from the custom errors? I don't think contract size is an issue here.
it would also reduce the diff and external interface compared to AuctionHouseV1

wdyt?

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,428 @@
...
+import { ReentrancyGuardUpgradeable } from '@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol';
+import { OwnableUpgradeable } from '@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol';
+import { IERC20 } from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import { INounsAuctionHouseV2 } from './interfaces/INounsAuctionHouseV2.sol';
+import { INounsToken } from './interfaces/INounsToken.sol';
+import { IWETH } from './interfaces/IWETH.sol';
+
+contract NounsAuctionHouseV2 is
+    INounsAuctionHouseV2,
+    PausableUpgradeable,
```
davidbrai, line 37: important to document that `Pausable` and `ReentrancyGuard` are only used for taking up storage slots

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,428 @@
...
+
+        reservePrice = _reservePrice;
+        timeBuffer = _timeBuffer;
+        minBidIncrementPercentage = _minBidIncrementPercentage;
+    }
+
+    /**
+     * @notice Settle the current auction, mint a new Noun, and put it up for auction.
+     */
+    function settleCurrentAndCreateNewAuction() external override whenNotPaused {
```
davidbrai, line 106: @eladmallel `whenNotPaused` and `whenPaused` are not using the new `__paused` variable

davidbrai: ignore!

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /**
+     * @notice Settle an auction, finalizing the bid and paying out to the owner.
+     * @dev If there are no bids, the Noun is burned.
+     */
+    function _settleAuction() internal {
+        INounsAuctionHouseV2.AuctionV2 memory _auction = auction;
+
+        if (_auction.startTime == 0) revert AuctionHasntBegun();
```
davidbrai, line 250: same comment here regarding custom errors vs string errors

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,428 @@
...
+        if (!auction.settled && latestNounId > 0) {
+            latestNounId -= 1;
+        }
+
+        settlements = new Settlement[](auctionCount);
+        uint256 actualCount = 0;
+        while (actualCount < auctionCount && latestNounId > 0) {
+            // Skip Nouner reward Nouns, they have no price
+            // Also skips IDs with no price data
+            if (settlementHistory[latestNounId].blockTimestamp == 0) {
```
davidbrai, line 354: @eladmallel if someone called `warmUpSettlementState`, then the `blockTimestamp` will be `== 1` and this check will be false.
so we need to check `<= 1`, or look at `winner` or `amount`
I'm wondering if we need to keep this warmUp function at all.

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,428 @@
...
+
+    /// @notice The minimum percentage difference between the last bid amount and the current bid
+    uint8 public minBidIncrementPercentage;
+
+    /// @notice The active auction
+    INounsAuctionHouseV2.AuctionV2 public auction;
+
+    /// @notice Whether this contract is paused or not
+    /// @dev Replaces the state variable from PausableUpgradeable, to bit pack this bool with `auction` and save gas
+    bool public __paused;
```
davidbrai, line 67: @eladmallel I don't this is actually packs it in storage:
> Items following struct or array data always start a new storage slot.

from: https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html

davidbrai: you can run the build with storage layout output to see it also:

```
      {
         "astId":50,
         "contract":"contracts/NounsAuctionHouseV2.sol:NounsAuctionHouseV2",
         "label":"auction",
         "offset":0,
         "slot":"202",
         "type":"t_struct(AuctionV2)1382_storage"
      },
      {
         "astId":53,
         "contract":"contracts/NounsAuctionHouseV2.sol:NounsAuctionHouseV2",
         "label":"__paused",
         "offset":0,
         "slot":"204",
         "type":"t_bool"
      },
```

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: fixed: 05eb47a49affc06cef39675a5d80c663ae48df03

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/799/commits/79f611a6de8323c8ba7e1063087a4c99eb3e4a80

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/799/commits/79f611a6de8323c8ba7e1063087a4c99eb3e4a80

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: fixed

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: (copy pasting comment from discord)
placing it after the `AuctionV2` struct puts it in a new storage slot.
to gain gas savings, it needs to be in the same slot as the `AuctionV2`, because only that is accessed during `settleCurrentAndCreateNewAuction`.

I'm apprehensive to add it to the struct because:
1. it's not logically part of that struct
2. we'll need to change the `auction()` view function to return a struct without paused for backwards compatibility.

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,520 @@
...
+        for (uint256 i = 0; i < nounIds.length; ++i) {
+            if (settlementHistory[nounIds[i]].blockTimestamp == 0) {
+                settlementHistory[nounIds[i]] = SettlementState({ blockTimestamp: 1, amount: 0, winner: address(0) });
+            }
+        }
+    }
+
+    /**
+     * @notice Get past auction settlements.
+     * @dev Returns settlements in reverse order, meaning settlements[0] will be the most recent auction price.
```
eladmallel, line 357: @davidbrai maybe worth updating all the natspecs of the view functions, to say that they are skipping burned auctions?
(it wasn't the case until you changed the skip logic to when winner is address zero)

davidbrai: @eladmallel good point. I hadn't considered it. but I think it's more correct not to include these as they don't represent an actual sold noun

PR #800, opened 2023-09-28
The Burn: spend or burn excess ETH

eladmallel:


packages/nouns-contracts/contracts/governance/ExcessETH.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
+     *   PUBLIC FUNCTIONS
+     * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
+     */
+
+    function excessETH() public view returns (uint256) {
+        uint256 expectedTreasuryValue = meanAuctionPrice() * dao.adjustedTotalSupply();
+        return min(treasuryValueInETH() - expectedTreasuryValue, owner().balance);
```
solimander, line 101: Perhaps throw a custom error instead if there is there is no excess ETH.

eladmallel: this min is meant to allow us to burn as much as we can; eg excess is 3K ETH but we only have 2K in native ETH, let's at least burn that.

for the case of no excess I agree we should just not burn, will add a condition for that in the treasury contract

packages/nouns-contracts/contracts/governance/ExcessETH.sol
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * @notice Get the current treasury value denomiated in ETH.
+     * In addition to native ETH, it also includes stETH, wETH, and rETH balances.
+     * @dev for rETH, it uses the getEthValue() function to convert rETH to ETH.
+     */
+    function treasuryValueInETH() public view returns (uint256) {
+        address owner_ = owner();
+        return owner_.balance + stETH.balanceOf(owner_) + wETH.balanceOf(owner_) + rETHBalanceInETH();
```
davidbrai, line 142: @eladmallel nit: feels cleaner to pass `_owner` to `rETHBalanceInETH`

packages/nouns-contracts/test/foundry/governance/NounsDAOExecutorV3.t.sol
```diff
@@ -0,0 +1,219 @@
...
+        treasury.setExcessETHHelper(newExcessETH);
+
+        assertEq(address(treasury.excessETHHelper()), address(newExcessETH));
+    }
+
+    function test_burnExcessETH_givenHelperNotSet_reverts() public {
+        vm.prank(address(treasury));
+        treasury.setExcessETHHelper(IExcessETH(address(0)));
+
+        vm.expectRevert(abi.encodeWithSelector(NounsDAOExecutorV3.ExcessETHHelperNotSet.selector));
```
davidbrai, line 64: @eladmallel nit: you don't need to wrap with `abi.encodeWithSelector`.
this:
`vm.expectRevert(NounsDAOExecutorV3.ExcessETHHelperNotSet.selector);`
should work

eladmallel: thanks, fixed!

packages/nouns-contracts/contracts/governance/ExcessETH.sol
eladmallel: thanks, fixed!

PR #803, opened 2023-10-02
The Burn: spend or burn excess ETH, every N Noun mints

eladmallel:


packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
```diff
@@ -0,0 +1,247 @@
...
+    INounsAuctionHouseV2 public immutable auction;
+    IERC20 public immutable wETH;
+    IERC20 public immutable stETH;
+    IERC20 public immutable rETH;
+    uint128 public nextBurnNounID;
+    uint128 public minNewNounsBetweenBurns;
+    uint16 public numberOfPastAuctionsForMeanPrice;
+
+    constructor(
+        address owner_,
```
davidbrai, line 86: nit: let's make sure to add natspec here that states that `owner_` is expected to be a timelock, implementing `IExecutorV3`

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
```diff
@@ -0,0 +1,247 @@
...
+        return owner_.balance + stETH.balanceOf(owner_) + wETH.balanceOf(owner_) + rETHBalanceInETH(owner_);
+    }
+
+    /**
+     * @notice Get the mean auction price of the last `numberOfPastAuctionsForMeanPrice` auctions.
+     * @dev Reverts if there is not enough auction history to calculate the mean auction price.
+     */
+    function meanAuctionPrice() public view returns (uint256) {
+        uint16 numberOfPastAuctionsForMeanPrice_ = numberOfPastAuctionsForMeanPrice;
+        INounsAuctionHouseV2.Settlement[] memory settlements = auction.prices(numberOfPastAuctionsForMeanPrice_);
```
davidbrai, line 176: @eladmallel maybe we should add a function that returns only the prices instead the entire settlement structs.
it will take up less gas, and may be useful for other downstream applications

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
```diff
@@ -0,0 +1,247 @@
...
+     * @notice Get the amount of excess ETH in the Nouns treasury.
+     * Excess ETH is defined as the difference between the current treasury value denominated in ETH,
+     * and the expected treasury value denominated in ETH, which is defined as mean auction price * adjusted total supply.
+     * It returns zero during the waiting period, and when expected treasury value is greater than current value.
+     * If treasury balance in native ETH is less than the excess ETH calculation, it returns the treasury ETH balance.
+     * @dev This version does not support burning excess other than in native ETH, e.g. stETH, rETH, etc.
+     * @dev Reverts if there is not enough auction history to calculate the mean auction price.
+     */
+    function excessETH() public view returns (uint256) {
+        uint256 expectedTreasuryValue = expectedTreasuryValueInETH();
```
davidbrai, line 144: nit: maybe a better name would be `allowedTreasuryValue`?

packages/nouns-contracts/contracts/governance/NounsDAOExecutorV3.sol
```diff
@@ -0,0 +1,286 @@
...
+pragma solidity ^0.8.19;
+
+import { IERC20 } from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import { SafeERC20 } from '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+import { Initializable } from '@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol';
+import { UUPSUpgradeable } from '@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol';
+import { Address } from '@openzeppelin/contracts/utils/Address.sol';
+import { Burn } from '../libs/Burn.sol';
+
+contract NounsDAOExecutorV3 is UUPSUpgradeable, Initializable {
```
davidbrai, line 49: @eladmallel I'm not sure if it's better, but wanted your opinion.

I think we can make `NounsDAOExecutorV3` inherit from `NounsDAOExecturoV2` and only add the new burn related functions.
wdyt?

packages/nouns-contracts/contracts/interfaces/IExcessETHBurner.sol
```diff
@@ -0,0 +1,22 @@
...
+ * ░░██░░██░░░████░░██░░░████░░░ *
+ * ░░██░░██░░░████░░██░░░████░░░ *
+ * ░░░░░░█████████░░█████████░░░ *
+ * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ *
+ * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ *
+ *********************************/
+
+pragma solidity ^0.8.19;
+
+interface IExcessETHBurner {
```
davidbrai, line 20: @eladmallel maybe we can delete this?

packages/nouns-contracts/script/executorV3AndExcessETHBurner/DeployExecutorV3AndExcessETHBurnerBase.s.sol
```diff
@@ -0,0 +1,65 @@
...
+        address wETH_,
+        address stETH_,
+        address rETH_,
+        uint128 burnStartNounID_,
+        uint128 minNewNounsBetweenBurns_,
+        uint16 numberOfPastAuctionsForMeanPrice_
+    ) {
+        executorProxy = executorProxy_;
+
+        daoProxy = NounsDAOLogicV3(payable(NounsDAOExecutorV3(executorProxy_).admin()));
```
davidbrai, line 33: @eladmallel to reduce compilation time, it's better to not import `NounsDAOLogicV3` but instead to use an interface when making calls.
otherwise, whenever we make a change to `NounsDAOLogicV3` then this file will also need to be recompiled.

packages/nouns-contracts/script/executorV3AndExcessETHBurner/ProposeExecutorV3UpgradeBase.s.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.15;
+
+import 'forge-std/Script.sol';
+import { NounsDAOLogicV3 } from '../../contracts/governance/NounsDAOLogicV3.sol';
```
davidbrai, line 5: @eladmallel same comment as above regarding using an interface instead of importing `NounsDAOLogicV3`

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
davidbrai: fixed: 23b84bafa3f9b0105100cb7ac9b7aa0a42d4adbe

packages/nouns-contracts/script/executorV3AndExcessETHBurner/ProposeExecutorV3UpgradeBase.s.sol
```diff
@@ -0,0 +1,44 @@
...
+    function propose(
+        address[] memory targets,
+        uint256[] memory values,
+        string[] memory signatures,
+        bytes[] memory calldatas,
+        string memory description
+    ) external returns (uint256);
+}
+
+abstract contract ProposeExecutorV3UpgradeBase is Script {
```
eladmallel, line 16: @davidbrai do you want to update this script to do more stuff, to basically align it with your mainnet fork test?

would be good to use as many scripts as possible as part of that flow

packages/nouns-contracts/test/foundry/BurnUpgradeForkMainnetTest.t.sol
```diff
@@ -0,0 +1,264 @@
...
+
+        // treasury has less ETH than 13899 - 25 * 422
+        assertEq(address(NOUNS_TIMELOCK_V2_MAINNET).balance, 2813.062865210366418892 ether);
+
+        burner.burnExcessETH();
+
+        assertEq(address(NOUNS_TIMELOCK_V2_MAINNET).balance, 0);
+    }
+
+    function proposeToTurnOnBurn(
```
eladmallel, line 232: @davidbrai maybe this should also become a script?

packages/nouns-contracts/test/foundry/governance/ExcessETHBurner.t.sol
```diff
@@ -0,0 +1,281 @@
...
+    function test_burnExcessETH_givenInsufficientAuctionHistory_reverts() public {
+        vm.deal(address(treasury), 100 ether);
+        dao.setAdjustedTotalSupply(1);
+        setPriceHistory(1 ether, pastAuctionCount - 1);
+
+        vm.expectRevert(ExcessETHBurner.NotEnoughAuctionHistory.selector);
+        burner.burnExcessETH();
+    }
+
+    // setMinNewNounsBetweenBurns
```
eladmallel, line 209: a little leftover :)

davidbrai: fixed

packages/nouns-contracts/test/foundry/BurnUpgradeForkMainnetTest.t.sol
davidbrai: @eladmallel I've decided not to make the proposals a script because I think we will make the proposal manually.
it's not a complex proposal like the V3 upgrade, so I think it's safe.
wdyt?

packages/nouns-contracts/script/executorV3AndExcessETHBurner/ProposeExecutorV3UpgradeBase.s.sol
davidbrai: @eladmallel see my reply below regarding not doing the proposal part via script

packages/nouns-contracts/test/foundry/BurnUpgradeForkMainnetTest.t.sol
```diff
@@ -0,0 +1,264 @@
...
+            vm.warp(ah.auction().endTime + 1);
+            ah.settleCurrentAndCreateNewAuction();
+        }
+
+        assertEq(burner.meanAuctionPrice(), 25 ether);
+        assertEq(NOUNS_DAO_PROXY_MAINNET.adjustedTotalSupply(), 422);
+        assertEq(burner.expectedTreasuryValueInETH(), 10550 ether);
+        assertEq(burner.treasuryValueInETH(), 13899.047817579079479237 ether);
+
+        // treasury has less ETH than 13899 - 25 * 422
```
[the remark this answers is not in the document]
davidbrai: @eladmallel added a comment

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
```diff
@@ -125,14 +126,16 @@ contract ExcessETHBurner is Ownable {
...
+        if (currentNounId < nextBurnNounID_) revert NotTimeToBurnYet();
 
         amount = excessETH();
         if (amount == 0) revert NoExcessToBurn();
 
         IExecutorV3(owner()).burnExcessETH(amount);
 
-        nextBurnNounID += minNewNounsBetweenBurns;
+        nextBurnNounID = nextBurnNounID_ + minNewNounsBetweenBurns;
+        emit Burn(amount, nextBurnNounID_, nextBurnNounID);
```
eladmallel, line 149: names of variables here feel confusing, and there's a gas nit here of reading `nextBurnNounID` from storage right after calculating its value in this function.

maybe store the new value in a local variable like `newNextBurnNounID`?
maybe change `nextBurnNounID_` to `nextBurnNounIDCache` to further signal it's caching the previous value?

davidbrai: fixed: ea002b57cb3df093d5fc593678896977c46a021c

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,511 @@
...
+     * @dev Only writes to slots where blockTimestamp is zero, meaning it will not overwrite existing data.
+     * @param nounIds The list of Noun IDs whose settlement slot to warm up.
+     */
+    function warmUpSettlementState(uint256[] calldata nounIds) external {
+        for (uint256 i = 0; i < nounIds.length; ++i) {
+            if (settlementHistory[nounIds[i]].blockTimestamp == 0) {
+                settlementHistory[nounIds[i]] = SettlementState({ blockTimestamp: 1, amount: 0, winner: address(0) });
+            }
+        }
+    }
```
solimander, line 340: It's likely worth optimizing this function since it may be called to warm up thousands of slots.

e.g.
```solidity
function warmUpSettlementState(uint256[] calldata nounIds) external {
    uint256 nounIdCount = nounIds.length;
    SettlementState storage settlementState;
    for (uint256 i; i < nounIdCount;) {
        settlementState = settlementHistory[nounIds[i]];
        if (settlementState.blockTimestamp == 0) {
            settlementState.blockTimestamp = 1;
        }
        unchecked {
            ++i;
        }
    }
}
```

solimander, line 340: It's likely worth optimizing this function since it may be called to warm up thousands of slots.

e.g.
```solidity
function warmUpSettlementState(uint256[] calldata nounIds) external {
    uint256 nounIdCount = nounIds.length;
    SettlementState storage settlementState;
    for (uint256 i; i < nounIdCount;) {
        settlementState = settlementHistory[nounIds[i]];
        if (settlementState.blockTimestamp == 0) {
            settlementState.blockTimestamp = 1;
        }
        unchecked {
            ++i;
        }
    }
}
```

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,511 @@
...
+        prices = new uint256[](auctionCount);
+        uint256 actualCount = 0;
+        while (actualCount < auctionCount && latestNounId > 0) {
+            SettlementState memory settlementState = settlementHistory[latestNounId];
+            // Skip Nouner reward Nouns, they have no price
+            // Also skips IDs with no price data
+            if (settlementState.winner == address(0)) {
+                --latestNounId;
+                continue;
+            }
```
solimander, line 407: I'm not so sure that we should skip auctioned nouns with no price data. Any gaps in historical population may return unexpected price results that could in turn affect the burn.

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,511 @@
...
+     */
+    function getPrices(uint256 auctionCount) external view returns (uint256[] memory prices) {
+        uint256 latestNounId = auctionStorage.nounId;
+        if (!auctionStorage.settled && latestNounId > 0) {
+            latestNounId -= 1;
+        }
+
+        prices = new uint256[](auctionCount);
+        uint256 actualCount = 0;
+        while (actualCount < auctionCount && latestNounId > 0) {
```
solimander, line 400: May be worth optimizing this loop as it could run 90+ times in the proposed design.

e.g.

```solidity
SettlementState memory settlementState;
uint256 actualCount;
for (uint256 id = latestNounId; id > 0 && actualCount < auctionCount; --id) {
    if (id % 10 == 0 && id <= 1820) continue; // Skip Nounder nouns.

    settlementState = settlementHistory[id];
    if (settlementState.winner == address(0)) continue; // Skip IDs with no price data.

    prices[actualCount] = uint64PriceToUint256(settlementState.amount);
    unchecked {
        ++actualCount;
    }
}
```

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    /**
+     * @notice Get the current treasury value denomiated in ETH.
+     * In addition to native ETH, it also includes stETH, wETH, and rETH balances.
+     * @dev for rETH, it uses the getEthValue() function to convert rETH to ETH.
+     */
+    function treasuryValueInETH() public view returns (uint256) {
+        address owner_ = owner();
+        return owner_.balance + stETH.balanceOf(owner_) + wETH.balanceOf(owner_) + rETHBalanceInETH(owner_);
```
solimander, line 237: Thoughts on including token buyer's ETH in this calculation as well? 10% of the treasury currently sits in the token buyer.

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    /**
+     * @notice Get the mean auction price of the last `numberOfPastAuctionsForMeanPrice` auctions.
+     * @dev Reverts if there is not enough auction history to calculate the mean auction price.
+     */
+    function meanAuctionPrice() public view returns (uint256) {
```
solimander, line 244: May be worth adding a mean view function to the auction house itself to remove the need to re-loop over all prices. It also feels like it may be desirable for others who want to build on Nouns.

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
```diff
@@ -0,0 +1,323 @@
...
+     * @notice Sets `nounIdsBetweenBurns`
+     * Can only be called by owner, which is assumed to be the NounsDAOExecutorV3 contract, i.e. the Nouns treasury.
+     */
+    function setNounIdsBetweenBurns(uint64 newNounIdsBetweenBurns) external onlyOwner {
+        emit NounIdsBetweenBurnsSet(nounIdsBetweenBurns, newNounIdsBetweenBurns);
+
+        nounIdsBetweenBurns = newNounIdsBetweenBurns;
+    }
+
+    function setBurnWindowSize(uint16 newBurnWindowSize) external onlyOwner {
```
solimander, line 292: nit: Add comment

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: let's discuss this

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
davidbrai: let's discuss this.
to consider: token buyer's ETH is not part of the forked assets, so it doesn't participate in the BV

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: fixed: 5d1bea64af093f87459c1cd7bfc28556173b8e7a

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: @solimander At first I thought you meant nouns that are auctioned but got no bids, and therefore burned.
But now I think I understand you meant historical data gaps for which we didn't populate data.

Trying to build a more concrete example:
* We deploy this and start saving data starting from noun 900
* We populate past data for nouns 800-850
* Latest noun is at 925 and another contract calls `getPrices(50)`, then they get the prices for nouns 825-850 & 900-925 (ignoring nounder rewards for now), which is not what they're expecting.

1. Is this what you meant? if not, could you provide an example that demonstrates the issue?
2. Thinking what would be a better behavior in this case. Some options:
  a. Don't skip the ids, but also don't return data for them. e.g. for `getPrices(50)`, returns only prices for nouns 900-925. for `getPrices(100)` it would return prices for nouns 825-850 and 900-925.
  b. Return 0 as the price for ids with no data. e.g. `getPrices(50)` would get prices for nouns 900-925 and 25 zeroes. Then the caller of the function would need to decide what to do with zeroes.
  c. Any other ideas?
3. In both suggestions above, my understanding is that you think we should skip nounder rewards nouns, is that correct? To do that I think we would need to add the `id <= 1820 && id % 10 == 0` logic here, right?

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
davidbrai: @solimander we haven't considered including token buyer's ETH in the calculation so far.
agree that currently a big chunk of the treasury is there, but I think that's an anomaly rather than the norm. We're got to much ETH sent there because many were looking at the state of token buyer as a candidate and those transactions haven't since updated. I feel like this is a more of a bug with token buyer rather than a normal state. Perhaps token buyer should send ETH back to the DAO if it has too much? haven't thought about it enough.

Regardless, I think there's some logic in excluding it from the calculation here, because it doesn't participate in the assets sent to a fork. I don't feel strongly about it, so curious to hear your thoughts here.
also @eladmallel

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
solimander: > 1. Is this what you meant? if not, could you provide an example that demonstrates the issue?

Yes, that's the case I'm referring to.

> 2. Thinking what would be a better behavior in this case. Some options:
a. Don't skip the ids, but also don't return data for them. e.g. for getPrices(50), returns only prices for nouns 900-925. for getPrices(100) it would return prices for nouns 825-850 and 900-925.
b. Return 0 as the price for ids with no data. e.g. getPrices(50) would get prices for nouns 900-925 and 25 zeroes. Then the caller of the function would need to decide what to do with zeroes.
c. Any other ideas?

I think that's an improvement on the existing logic. Alternatively, we could revert if there aren't enough consecutive prices populated to service the query (excluding Nounder nouns). Perhaps we expose both and let the consumer decide which fits their use-case.

> 3. In both suggestions above, my understanding is that you think we should skip nounder rewards nouns, is that correct? To do that I think we would need to add the id <= 1820 && id % 10 == 0 logic here, right?

Yes, that's right.

packages/nouns-contracts/contracts/governance/ExcessETHBurner.sol
solimander: Okay, makes sense. I agree it likely doesn't make sense to add since it's not included in BV, but we should probably address this issue at some point.

eladmallel: I thought about it for a brief moment some weeks ago and decided to leave it out for the reason you mentioned - since that ETH is anyway currently not part of the forking mechanism.

but since it has the potential to hold significant amounts of ETH, and has the potential to be used to DAO spending, it can be a weird way of gaming the burn against our intentions, and because it seems fairly simple to take it into account - I am now leaning towards including its ETH in this calculation.

@davidbrai let's just add it?

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
davidbrai: Another case to consider:
A noun auctioned that got no bids, I think it this case it *does* make sense to skip, because it's not missing data.

solimander: Thoughts on the following? -

```solidity
function getPrices(uint256 auctionCount) external view returns (uint256[] memory prices) {
    uint256 latestNounId = auctionStorage.nounId;
    if (!auctionStorage.settled && latestNounId > 0) {
        latestNounId -= 1;
    }

    prices = new uint256[](auctionCount);
    SettlementState memory settlementState;
    uint256 actualCount;
    for (uint256 id = latestNounId; id > 0 && actualCount < auctionCount; --id) {
        if (id % 10 == 0 && id <= 1820) continue; // Skip Nounder nouns.

        settlementState = settlementHistory[id];
        if (settlementState.blockTimestamp <= 1) revert NOT_ENOUGH_CONSECUTIVE_PRICES();

        prices[actualCount] = uint64PriceToUint256(settlementState.amount);
        unchecked {
            ++actualCount;
        }
    }

    if (auctionCount > actualCount) {
        // this assembly trims the observations array, getting rid of unused cells
        assembly {
            mstore(prices, actualCount)
        }
    }
}
```

davidbrai: @solimander although unlikely, in case an auction doesn't get any bids and eventually the noun is burned, I think it's would be incorrect to assume the price there is zero. Since there was no actual transaction, I think the correct thing would be to skip it.
example: maybe there are no bids cause for some reason ethereum was down or gas prices were extreme.
wdyt?

solimander: Unlikely, but I see your point. Perhaps a more likely case is when a reserve price has been set and is not met in an auction.

PR #806, opened 2023-10-10
DRAFT: require noun age to fork

davidbrai:


packages/nouns-contracts/contracts/governance/fork/NounsDAOV3Fork.sol
```diff
@@ -216,6 +220,11 @@ library NounsDAOV3Fork {
         return ds.forkEscrow.numTokensInEscrow();
     }
 
+    function checkNounIdIsAllowedToFork(NounsDAOStorageV3.StorageV3 storage ds, uint256 tokenId) internal view {
+        uint256 auctionedNounId = INounsAuctionHouseV2(ds.nouns.minter()).auction().nounId;
+        if (tokenId < auctionedNounId - ds.nounAgeRequiredToFork) revert NounIdNotOldEnough();
```
davidbrai, line 225: gas: worth reading `ds.nounAgeRequiredToFork` and `auctionedNounId` once for all the tokens

packages/nouns-contracts/contracts/governance/fork/NounsDAOV3Fork.sol
```diff
@@ -221,6 +221,8 @@ library NounsDAOV3Fork {
     }
 
     function checkNounIdIsAllowedToFork(NounsDAOStorageV3.StorageV3 storage ds, uint256 tokenId) internal view {
+        if (ds.nounAgeRequiredToFork == 0) return;
```
davidbrai: @eladmallel I think if the age required is zero, it'll anyway pass

eladmallel: @davidbrai I think without this added line the code doesn't work. In this critical line:
`if (tokenId < auctionedNounId - ds.nounAgeRequiredToFork) revert NounIdNotOldEnough();`

if `nounAgeRequiredToFork` is zero, the condition is always true for all nouns that aren't currently on auction -- meaning no Noun can fork. All circulating Noun IDs would evaluate true and revert is executed.

I only added this line after running tests and seeing them failing due to this new revert.

Does this make sense?

davidbrai: @eladmallel I think that means this condition is incorrect
should be:
```
age = auctionedNounId - tokenId;
if (age < ageRequiredToFork) revert();
```

davidbrai: I see that you fixed it now: https://github.com/nounsDAO/nouns-monorepo/pull/806/commits/9c9b15ee0b90e76bcfb8aaecb1738a0e177386d9
marking as resolved

PR #810, opened 2023-10-26
fix error in locales

GoodDaisy:


volkyeth:
those translations were commented out, so no impact on the app.
ty anyway

PR #813, opened 2023-11-07
DAO V3.1: fix vote gas refund

eladmallel:
we accidentally used the old V2 code that refunds msg.sender it should be tx.origin
to refund gas payers of multisigs


packages/nouns-contracts/test/governance/NounsDAO/V3/voteRefund.test.ts
```diff
@@ -0,0 +1,384 @@
...
+const LONG_REASON = realLongReason + realLongReason;
+const REFUND_ERROR_MARGIN = ethers.utils.parseEther('0.001');
```
eladmallel, line 32: FYI @solimander @davidbrai I once again made the test less sensitive; it used to be `0.00015` and the latest code fails to reach this level of accuracy, I'm guessing because of the changes we did around voting in V3, and I don't think it's important to deep dive into this. wdyt?

davidbrai:
@eladmallel can we update the original `voteRefund.test.ts` to test against DAOV3 instead of copy pasting it?
now that we have tagged versions, I don't think it's helping us to have tests that test the V2 contract

packages/nouns-contracts/test/descriptor.test.ts
```diff
@@ -1,105 +0,0 @@
...
-
-  beforeEach(async () => {
-    snapshotId = await ethers.provider.send('evm_snapshot', []);
-  });
-
-  afterEach(async () => {
-    await ethers.provider.send('evm_revert', [snapshotId]);
-  });
-
-  it('should generate valid token uri metadata when data uris are disabled', async () => {
```
davidbrai, line 53: @eladmallel I didn't find an equivalent test in foundry

davidbrai:
@eladmallel LGTM

packages/nouns-contracts/test/descriptor.test.ts
eladmallel: decided it's not important to keep Descriptor tests when we are live with V2 (this is testing V1).

solimander (approved):

PR #816, opened 2023-11-16
Data: remove candidate creation fee when used to update a proposal

eladmallel:


packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
```diff
@@ -117,7 +132,18 @@ contract NounsDAOData is OwnableUpgradeable, UUPSUpgradeable, NounsDAODataEvents
         string memory slug,
         uint256 proposalIdToUpdate
     ) external payable {
-        if (!isNouner(msg.sender) && msg.value < createCandidateCost) revert MustBeNounerOrPaySufficientFee();
+        if (proposalIdToUpdate > 0) {
+            INounsDAO dao = INounsDAO(nounsDao);
+            if (dao.state(proposalIdToUpdate) != NounsDAOStorageV3.ProposalState.Updatable)
```
davidbrai, line 137: @eladmallel for gas purposes, maybe instead of calling `dao.state()`, you can check the `updatePeriodEndBlock` returned in the `propInfo` below?

packages/nouns-contracts/src/index.ts
```diff
@@ -14,5 +14,5 @@ export { NounsSeeder__factory as NounsSeederFactory } from '../typechain/factori
 export { NounsDAOLogicV1__factory as NounsDaoLogicV1Factory } from '../typechain/factories/contracts/governance/NounsDAOLogicV1__factory';
 export { NounsDAOLogicV2__factory as NounsDaoLogicV2Factory } from '../typechain/factories/contracts/governance/NounsDAOLogicV2__factory';
 export { NounsDAOLogicV3__factory as NounsDaoLogicV3Factory } from '../typechain/factories/contracts/governance/NounsDAOLogicV3__factory';
-export { NounsDAOData__factory as NounsDaoDataFactory } from '../typechain/factories/contracts/governance/data/NounsDAOData__factory';
+export { NounsDAOData__factory as NounsDaoDataFactory } from '../typechain/factories/contracts/governance/data/NounsDAOData.sol/NounsDAOData__factory';
```
davidbrai, line 17: @eladmallel is this change on purpose?

eladmallel: @davidbrai it's the result of adding `INounsDAO` in the data contract file. I actually like having these small interfaces that solve a specific need rather than the big interfaces, so I am ok with this change.

I think it's overkill to move this small interface to another file. but open to it if you prefer.
wdyt?

packages/nouns-contracts/contracts/governance/data/NounsDAOData.sol
eladmallel: I agree, pushed a fix

packages/nouns-contracts/src/index.ts
davidbrai: @eladmallel I don't mind the interface, was just surprised by the change in this file. guess it's fine!

PR #818, opened 2023-11-24
candidate signature expiration date in seconds

ripe0x:
- [x] send candidate signature expiration in seconds instead of milliseconds
- [x] update signature components to render expiration date using seconds
- [x] simplify candidates logic in multiple components by moving more data parsing logic into hook


packages/nouns-webapp/src/pages/EditCandidate/index.tsx
```diff
@@ -88,6 +97,8 @@ const EditCandidatePage: React.FC<EditCandidateProps> = props => {
     [proposalTransactions, totalUSDCPayment],
   );
 
+  console.log('candidate', candidate);
```
solimander, line 100: nit: remove log

PR #826, opened 2024-01-03
Client Incentives + Noracle

eladmallel:


packages/nouns-contracts/contracts/NounsAuctionHouse.sol
eladmallel, line 1: @davidbrai I think we can revert all changes done in this file and focus only on V2, wdyt?

packages/nouns-contracts/contracts/governance/NounsDAOV3Admin.sol
```diff
@@ -209,29 +206,29 @@ library NounsDAOV3Admin {
...
 
-        uint32 oldObjectionPeriodDurationInBlocks = ds.objectionPeriodDurationInBlocks;
-        ds.objectionPeriodDurationInBlocks = newObjectionPeriodDurationInBlocks;
+        uint32 oldObjectionPeriodDurationInBlocks = ds().objectionPeriodDurationInBlocks;
+        ds().objectionPeriodDurationInBlocks = newObjectionPeriodDurationInBlocks;
 
         emit ObjectionPeriodDurationSet(oldObjectionPeriodDurationInBlocks, newObjectionPeriodDurationInBlocks);
     }
 
+    function ds() internal pure returns (NounsDAOStorageV3.StorageV3 storage ds_) {
```
eladmallel, line 219: prefer to move this function to the bottom of the file

packages/nouns-contracts/contracts/governance/NounsDAOV3Votes.sol
```diff
@@ -146,7 +145,13 @@ library NounsDAOV3Votes {
         uint8 support,
         string calldata reason
     ) external {
-        emit VoteCast(msg.sender, proposalId, support, castVoteInternal(ds, msg.sender, proposalId, support), reason);
+        emit VoteCast(
```
eladmallel, line 148: I think it's useful to have an emit that includes the client ID. let's discuss whether we add it here or add a new event.
I know we've made sure to be compatible with GovernorBravo with these events so far, I'm just not convinced that's important anymore.

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
```diff
@@ -890,17 +909,19 @@ library NounsDAOV3Proposals {
...
+        ProposalTxs memory txs,
+        uint32 clientId
     ) internal returns (NounsDAOStorageV3.Proposal storage newProposal) {
         uint64 updatePeriodEndBlock = SafeCast.toUint64(block.number + ds.proposalUpdatablePeriodInBlocks);
         uint256 startBlock = updatePeriodEndBlock + ds.votingDelay;
         uint256 endBlock = startBlock + ds.votingPeriod;
 
         newProposal = ds._proposals[proposalId];
         newProposal.id = proposalId;
+        newProposal.clientId = clientId;
```
eladmallel, line 924: similar to the comment on cast vote, wouldn't it be cool to include the client ID in new proposal events? let's discuss

packages/nouns-contracts/contracts/interfaces/INounsAuctionHouseRewards.sol
eladmallel, line 1: Why do we need this interface vs just adding these functions to the V2 interface?

packages/nouns-contracts/contracts/interfaces/INounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,106 @@
...
+
+    function getPrices(uint256 auctionCount) external view returns (uint256[] memory prices);
+
+    function getSettlements(
+        uint256 startId,
+        uint256 endId,
+        bool skipEmptyValues
+    ) external view returns (Settlement[] memory settlements);
+
+    function warmUpSettlementState(uint256[] calldata nounIds) external;
```
eladmallel, line 103: capturing that we're questioning the warm up approach; it might be better to just do a gas refund when writing settlement history. it's cheaper in total gas and simpler technically.

packages/nouns-contracts/package.json
```diff
@@ -42,7 +42,7 @@
     "hardhat": "^2.9.7",
     "hardhat-abi-exporter": "^2.9.0",
     "hardhat-gas-reporter": "^1.0.8",
-    "prettier-plugin-solidity": "^1.0.0-beta.12",
+    "prettier-plugin-solidity": "^1.2.0",
```
eladmallel, line 45: curious why you chose to upgrade this lib?

packages/nouns-contracts/contracts/NounsAuctionHouse.sol
davidbrai: @eladmallel yes

packages/nouns-contracts/contracts/governance/NounsDAOV3Admin.sol
davidbrai: sure

packages/nouns-contracts/contracts/governance/NounsDAOV3Votes.sol
davidbrai: agree we should add the event
and we need to think about whether to keep compatibility for gov bravo

packages/nouns-contracts/contracts/governance/NounsDAOV3Proposals.sol
davidbrai: agree

packages/nouns-contracts/contracts/interfaces/INounsAuctionHouseRewards.sol
davidbrai: I don't remember why I added it. we can remove it

packages/nouns-contracts/contracts/NounsAuctionHouseV2.sol
```diff
@@ -0,0 +1,509 @@
...
+        );
+
+        auctionStorage.clientId = clientId;
+        auctionStorage.amount = uint128(msg.value);
+        auctionStorage.bidder = payable(msg.sender);
+
+        // Extend the auction if the bid was received within `timeBuffer` of the auction end time
+        bool extended = _auction.endTime - block.timestamp < _timeBuffer;
+
+        emit AuctionBid(_auction.nounId, msg.sender, msg.value, extended);
```
davidbrai, line 147: let's have an event that includes clientId also
perhaps extend this one?

packages/nouns-contracts/abi/contracts/governance/NounsDAOLogicV3.sol/NounsDAOLogicV3.json
```diff
@@ -44,30 +44,6 @@
     "name": "UnsafeUint16Cast",
     "type": "error"
   },
-  {
-    "inputs": [],
-    "name": "VotesBelowProposalThreshold",
```
davidbrai: @eladmallel there are a lot of changes in this file, do they all make sense?
maybe we should go over them

packages/nouns-contracts/abi/contracts/governance/NounsDAOLogicV4.json
```diff
@@ -0,0 +1,3164 @@
+[
```
davidbrai, line 1: maybe we should rename the contract to V4 then also

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
```diff
@@ -865,6 +566,39 @@ contract NounsDAOStorageV3 {
         bool executeOnTimelockV1;
     }
 
+    struct ProposalCondensedV2 {
```
davidbrai, line 569: @eladmallel the naming here is a bit confusing, because `ProposalCondensedV2` is actually older than `ProposalCondensed`, right?
any ideas how to make more intuitive?

packages/nouns-contracts/contracts/governance/NounsDAOProxyV3.sol
```diff
@@ -38,9 +38,21 @@
 
 pragma solidity ^0.8.19;
 
-import './NounsDAOInterfaces.sol';
+import { NounsDAOV3Types, NounsDAOV3Types } from './NounsDAOInterfaces.sol';
```
davidbrai, line 41: @eladmallel imported twice by accident?

packages/nouns-contracts/contracts/governance/fork/ForkDAODeployer.sol
```diff
@@ -18,11 +18,10 @@
 pragma solidity ^0.8.19;
 
 import { ERC1967Proxy } from '@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol';
-import { IForkDAODeployer, INounsDAOForkEscrow, NounsDAOStorageV3 } from '../NounsDAOInterfaces.sol';
```
davidbrai, line 21: @eladmallel I think the changes in this file can all be reverted?
optimizing for minimum diff on live contracts

packages/nouns-contracts/contracts/governance/fork/NounsDAOV3Fork.sol
```diff
@@ -17,7 +17,7 @@
 
 pragma solidity ^0.8.19;
 
-import { NounsDAOStorageV3, INounsDAOForkEscrow, INounsDAOExecutorV2 } from '../NounsDAOInterfaces.sol';
```
davidbrai, line 20: @eladmallel here maybe we can also revert the changes?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+    uint32 public nextTokenId = 1;
+
+    uint256 public lastProcessedAuctionId;
+    uint32 public nextProposalIdToReward;
+    uint256 public nextProposalRewardFirstAuctionId;
+    uint256 lastProposalRewardsUpdate = block.timestamp;
+
+    RewardParams public params;
+
+    IERC20 public ethToken; // TODO make configurable
```
eladmallel, line 38: minor gas: possible to put this token in the params struct and shave 8 bits off of the min reward period, so everything fits into a single slot

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+contract Rewards is ERC721('NounsClientIncentives', 'NounsClientIncentives'), Ownable {
+    INounsDAOLogicV3 public immutable nounsDAO;
+    INounsAuctionHouseV2 public immutable auctionHouse;
+
+    uint32 public nextTokenId = 1;
+
+    uint256 public lastProcessedAuctionId;
+    uint32 public nextProposalIdToReward;
+    uint256 public nextProposalRewardFirstAuctionId;
+    uint256 lastProposalRewardsUpdate = block.timestamp;
```
eladmallel, line 34: we can probably do some bit packing to all variables related to processing proposals, wdyt?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+
+    RewardParams public params;
+
+    IERC20 public ethToken; // TODO make configurable
+
+    mapping(uint32 clientId => uint256 balance) public clientBalances;
+
+    struct ClientData {
```
eladmallel, line 42: leftover?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+        address ethToken_,
+        uint32 nextProposalIdToReward_,
+        uint256 lastProcessedAuctionId_,
+        uint256 nextProposalRewardFirstAuctionId_,
+        RewardParams memory rewardParams
+    ) Ownable(owner) {
+        nounsDAO = INounsDAOLogicV3(nounsDAO_);
+        auctionHouse = INounsAuctionHouseV2(auctionHouse_);
+        ethToken = IERC20(ethToken_);
+        nextProposalIdToReward = nextProposalIdToReward_;
```
eladmallel, line 68: can we deduce this by reading the current prop ID from the DAO and maybe incrementing by one?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+        lastProcessedAuctionId = lastProcessedAuctionId_;
+        nextProposalRewardFirstAuctionId = nextProposalRewardFirstAuctionId_;
+        params = rewardParams;
+    }
+
+    struct Temp {
```
eladmallel, line 74: helpful to give this type a more specific name; and then either put it just above the function that uses it, or move it to some other place (e.g. a types interface) to declutter a bit, wdyt?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+        params = newParams;
+    }
+
+    function getParams() public view returns (RewardParams memory) {
+        return params;
+    }
+
+    function updateRewardsForAuctions(uint256 lastNounId) public {
+        uint256 currentlyAuctionedNounId = auctionHouse.auction().nounId;
+        require(lastNounId < currentlyAuctionedNounId, 'lastNounId must be settled');
```
eladmallel, line 90: the two requirements in this function mean that if we even stop auctions, the last auction can't be processed. I don't mind it so much, but if we take this approach I think it's good to add some dev natspecs explaining why you're doing this vs adding an OR condition allowing it to be the current noun on auction if it's settled = true.

wdyt?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+
+        uint256 lastProcessedAuctionId_ = lastProcessedAuctionId;
+        require(lastNounId > lastProcessedAuctionId_, 'lastNounId must be higher');
+        lastProcessedAuctionId = lastNounId;
+
+        INounsAuctionHouseV2.Settlement memory settlement;
+        for (uint256 nounId = lastProcessedAuctionId_ + 1; nounId <= lastNounId; nounId++) {
+            // TODO maybe can be optimized to one call with large range
```
eladmallel, line 98: probably a good idea, can save a bit of gas. curious why you preferred to use this approach in the first place?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+
+    function setParams(RewardParams calldata newParams) public onlyOwner {
+        params = newParams;
+    }
+
+    function getParams() public view returns (RewardParams memory) {
+        return params;
+    }
+
+    function updateRewardsForAuctions(uint256 lastNounId) public {
```
eladmallel, line 88: still planning to do gas refunds here, right?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+        uint256 lastProcessedAuctionId_ = lastProcessedAuctionId;
+        require(lastNounId > lastProcessedAuctionId_, 'lastNounId must be higher');
+        lastProcessedAuctionId = lastNounId;
+
+        INounsAuctionHouseV2.Settlement memory settlement;
+        for (uint256 nounId = lastProcessedAuctionId_ + 1; nounId <= lastNounId; nounId++) {
+            // TODO maybe can be optimized to one call with large range
+            settlement = auctionHouse.getSettlements(nounId, nounId + 1, false)[0];
+            if (settlement.clientId > 0) {
+                clientBalances[settlement.clientId] += (settlement.amount * params.auctionRewardBps) / 10_000;
```
eladmallel, line 101: a few days ago solidity 0.8.24 came out with tstore and tload. want to try and use that stuff to optimize gas here?

I'm thinking we can use tstore(p,v) to store per client ID their balance in this call, and then write to their clientBalances value only once in the end.

wdyt?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+            );
+        }
+    }
+
+    function withdrawClientBalance(uint32 clientId, uint256 amount, address to) public {
+        require(ownerOf(clientId) == msg.sender, 'must be client NFT owner');
+
+        clientBalances[clientId] -= amount;
+
+        ethToken.transfer(to, amount);
```
eladmallel, line 229: I think it's worthwhile to use `SafeERC20` for transfers, wdyt?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+        return sumAuctionsExcludingLast(s);
+    }
+
+    function sumAuctionsExcludingLast(INounsAuctionHouseV2.Settlement[] memory s) internal pure returns (uint256 sum) {
+        for (uint256 i = 0; i < s.length - 1; ++i) {
+            sum += s[i].amount;
+        }
+    }
+
+    function registerClient() public returns (uint32) {
```
eladmallel, line 263: let's remember to add some basic client metadata here
client URL seems like the bare minimum 
a display name could be nice as well

and wdyt about introducing a descriptor external dep to make it safer to render art in the future?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+        uint32 tokenId = nextTokenId++;
+        _mint(msg.sender, tokenId);
+        return tokenId;
+    }
+
+    function requireProposalEligibleForRewards(NounsDAOV3Types.ProposalCondensed memory proposal) internal view {
```
eladmallel, line 269: leftover?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+     * @param votingClientIds array of client ids that were used to vote on of all eligible the eligible proposals in
+     * this rewards distribution
+     */
+    function updateRewardsForProposalWritingAndVoting(
+        uint32 lastProposalId,
+        uint256 lastAuctionedNounId,
+        uint32[] calldata votingClientIds
+    ) public {
+        require(lastProposalId <= nounsDAO.proposalCount(), 'bad lastProposalId');
+        require(lastProposalId >= nextProposalIdToReward, 'bad lastProposalId');
```
eladmallel, line 120: worth caching `nextProposalIdToReward` into memory, I see it's currently read twice

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+     * The auction revenue up to this noun id are included in the bounty.
+     * @param votingClientIds array of client ids that were used to vote on of all eligible the eligible proposals in
+     * this rewards distribution
+     */
+    function updateRewardsForProposalWritingAndVoting(
+        uint32 lastProposalId,
+        uint256 lastAuctionedNounId,
+        uint32[] calldata votingClientIds
+    ) public {
+        require(lastProposalId <= nounsDAO.proposalCount(), 'bad lastProposalId');
```
eladmallel, line 119: why can't `lastProposalId` be the last proposal so far?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+            votingClientIds
+        );
+        nextProposalIdToReward = lastProposalId + 1;
+
+        NounsDAOV3Types.ProposalForRewards memory lastProposal = proposals[proposals.length - 1];
+
+        uint256 auctionRevenue = getAuctionRevenueBetweenNouns({
+            firstNounId: nextProposalRewardFirstAuctionId,
+            oneAfterLastNounId: lastAuctionedNounId,
+            endTimestamp: lastProposal.creationTimestamp
```
eladmallel, line 134: feels a bit awkward that `lastProposal` could be a proposal where creationTimestamp is zero. I believe the code will revert in `getAuctionRevenueBetweenNouns` because of the first require statement there.

maybe we should have a more explicit requirement that lastProposal has to have a non zero creation timestamp, wdyt?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+    struct ProposalRewardsParams {
+        uint32 lastProposalId;
+        uint256 expectedNumEligibleProposals;
+        uint256 expectedNumEligibleVotes;
+        uint256 firstNounId;
+        uint256 lastNounId;
+        uint32[] votingClientIds;
+    }
+
+    function getAuctionRevenueBetweenNouns(
```
eladmallel, line 241: wdyt of adding such a view function to AH? something like `getSettlements(firstNounId, endTimestamp)`?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+                clientBalances[settlement.clientId] += (settlement.amount * params.auctionRewardBps) / 10_000;
+            }
+        }
+    }
+
+    /**
+     * @param lastProposalId id of the last proposal to include in the rewards distribution. all proposals up to and
+     * including this id must have ended voting.
+     * @param lastAuctionedNounId the noun id that was auctioned when proposal with id `lastProposalId` was created.
+     * The auction revenue up to this noun id are included in the bounty.
```
eladmallel, line 110: worth adding explicitly that revenue for this noun is not included, e.g. "up to (not including)"

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,280 @@
...
+            }
+
+            uint256 votesInProposal;
+            NounsDAOV3Types.ClientVoteData[] memory voteData = proposals[i].voteData;
+            uint256 votes;
+            for (uint256 j; j < votingClientIds.length; ++j) {
+                clientId = votingClientIds[j];
+                votes = voteData[j].votes;
+                if (clientId != 0) {
+                    clientBalances[clientId] += votes * t.rewardPerVote;
```
eladmallel, line 213: tstore: here and in the proposal client ID reward above, might want to try using tstore/tload, doesn't feel like a must to me

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: this was created just to avoid stack-too-deep errors, so it has no real meaning.
but I will make sure to declutter

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: yup. removed

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: yeah we can do that.
what's the benefit? less parameters?

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: agree. haven't given this edge case much thought.

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: @eladmallel honestly I haven't polished this code yet. this was a quick & dirty version

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: @eladmallel thanks for the reminder. added to our TODOs. both here and for the proposal rewards

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: removed

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: @eladmallel not sure what you mean. I think it's allowed here...

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: maybe, though for it to happen we would need to initially set `nextProposalRewardFirstAuctionId` to a proposal before creation timestamp is being set. which we won't do.
still think it's helpful?

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: @eladmallel thinking about how it would work:
it could loop through all settled auction ids from `firstNounId` upwards, and stop when it reaches a timestamp > endTimestamp.
a bit annoying that it doesn't know the final array size in advance, but I'll give it a go. would be nice to remove the `lastNounId` parameter!

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: same comment as above

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
eladmallel: maybe a good solution is to rename the latest condensed struct to something like `ProposalCondensedV4`?

packages/nouns-contracts/contracts/Rewards.sol
eladmallel: yes, less room for human error :)

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: implemented!

packages/nouns-contracts/contracts/Rewards.sol
eladmallel: you're right, my bad

packages/nouns-contracts/contracts/Rewards.sol
eladmallel: yes, I am not sure why I somehow missed the equals part there

packages/nouns-contracts/contracts/Rewards.sol
eladmallel: let's clarify further what you have in mind; are you saying we will first deploy the DAO upgrade, then wait for the first proposal with the timestamp to be posted, and only then we will deploy the rewards contract?

davidbrai: @eladmallel yes, we'll first deploy the DAO upgrade & Noracle.
we need both proposals with timestamps and auctions with settlements.
once those are done, we deploy the rewards contract and and put a proposal to fund it either immediately or once it starts to accumulate some balances

packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -57,60 +62,95 @@ import {
...
-  proposal.onTimelockV1 = true;
-  proposal.save();
+export function handleProposalCreatedWithRequirementsV4(
+  event: ProposalCreatedWithRequirementsV4,
+): void {
+  handleParsedProposal(ParsedProposalV3.fromV4Event(event));
 }
 
-export function handleProposalCreated(parsedProposal: ParsedProposalV3): void {
+export function handleParsedProposal(parsedProposal: ParsedProposalV3): void {
```
davidbrai, line 144: would maybe rename `handleParsedProposal` to something like `saveProposalExtraDetails`

packages/nouns-subgraph/src/nouns-dao.ts
```diff
@@ -57,60 +62,95 @@ import {
...
-  proposal.lastUpdatedBlock = parsedProposal.createdBlock;
-  proposal.createdTransactionHash = parsedProposal.createdTransactionHash;
-  proposal.startBlock = parsedProposal.startBlock;
-  proposal.endBlock = parsedProposal.endBlock;
-  proposal.updatePeriodEndBlock = parsedProposal.updatePeriodEndBlock;
-  proposal.proposalThreshold = parsedProposal.proposalThreshold;
-  proposal.quorumVotes = parsedProposal.quorumVotes;
   proposal.forVotes = BIGINT_ZERO;
   proposal.againstVotes = BIGINT_ZERO;
   proposal.abstainVotes = BIGINT_ZERO;
```
davidbrai, line 149: is setting the votes to zero needed here? isn't it already done in `handleProposalCreated`?

packages/nouns-contracts/contracts/client-incentives/NounsClientToken.sol
```diff
@@ -0,0 +1,68 @@
...
+
+import { ERC721 } from '@openzeppelin/contracts-v5/token/ERC721/ERC721.sol';
+import { Ownable } from '@openzeppelin/contracts-v5/access/Ownable.sol';
+import { INounsClientTokenTypes } from './INounsClientTokenTypes.sol';
+import { INounsClientTokenDescriptor } from './INounsClientTokenDescriptor.sol';
+
+contract NounsClientToken is INounsClientTokenTypes, ERC721('Nouns Client Token', 'NOUNSCLIENT'), Ownable {
+    /// @dev This is a ERC-7201 storage location, calculated using:
+    /// @dev keccak256(abi.encode(uint256(keccak256("nounsclienttoken")) - 1)) & ~bytes32(uint256(0xff));
+    bytes32 private constant _STORAGE_LOCATION = 0x8cf5ce6e8ba000976223217bb8fd99e6473b9f0c4b7adc07d894a8f739887e00;
```
davidbrai, line 26: @eladmallel why not do `_STORAGE_LOCATION = keccak256(...)` ?

packages/nouns-contracts/contracts/client-incentives/NounsClientToken.sol
```diff
@@ -0,0 +1,68 @@
...
+                tokenId,
+                _getState().clientMetadata[uint32(tokenId)]
+            );
+    }
+
+    function setDescriptor(address descriptor) public onlyOwner {
+        _getState().descriptor = descriptor;
+    }
+
+    function STORAGE_LOCATION() external pure returns (bytes32) {
```
davidbrai, line 67: @eladmallel is this for convenience? you could make `STORAGE_LOCATION` public to auto create the getter

packages/nouns-contracts/contracts/client-incentives/INounsClientTokenTypes.sol
```diff
@@ -0,0 +1,29 @@
...
+
+pragma solidity ^0.8.19;
+
+interface INounsClientTokenTypes {
+    struct ClientMetadata {
+        string name;
+        string description;
+    }
+
+    struct TokenState {
```
davidbrai, line 24: maybe rename to `Storage` to indicate this is the entire storage of the token

packages/nouns-subgraph/src/nouns-dao.ts
eladmallel: done

packages/nouns-subgraph/src/nouns-dao.ts
eladmallel: because events are handled out of order on goldsky, we need to choose between setting them to zero in both function, vs making these values nullable as well. I chose to just set them again because we know they must be zero at this point. Making them nullable felt more annoying. wdyt?

packages/nouns-contracts/contracts/client-incentives/NounsClientToken.sol
eladmallel: if you do the calculation you get this compilation error:
```
Error (7615): Only direct number constants and references to such constants are supported by inline assembly.
```

Because we're using it in assembly to set the storage slot, it has to be a hard-coded value.

Makes sense?

packages/nouns-contracts/contracts/client-incentives/NounsClientToken.sol
eladmallel: done

packages/nouns-contracts/contracts/client-incentives/INounsClientTokenTypes.sol
eladmallel: done

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -320,6 +320,42 @@ contract Rewards is NounsClientToken, UUPSUpgradeable {
         return balance;
     }
 
+    /**
+     * @notice Returns the clientIds that is needed to be passed as a parameter to updateRewardsForProposalWritingAndVoting
+     * @dev This is not meant to be called onchain because it may be very gas intensive.
+     */
+    function getVotingClientIds(uint32 lastProposalId) public view returns (uint32[] memory) {
+        uint256 numClientIds = nextTokenId();
+        uint32[] memory allClientIds = new uint32[](numClientIds);
```
eladmallel, line 348: there's a risk someone might create so many client IDs that this function becomes very gas-intensive such that Etherscan can't run it.

I think it's not very interesting to mitigate this risk right now, but thought it good to discuss it here explicitly.

wdyt?

davidbrai: yes it's possible. I'm not so concerned since there's always the option of using an indexer to figure out the client ids that voted for a proposal

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
```diff
@@ -641,6 +641,6 @@ interface NounsDAOV3Types {
     }
 }
 
-contract NounsDAOStorageV3 is NounsDAOV3Types {
+contract NounsDAOStorage is NounsDAOV3Types {
     StorageV3 ds;
```
davidbrai, line 645: @eladmallel this part feels a bit off to me
still referencing V3 is many names
also, we did change the storage layout, so maybe it makes sense to change the name?

packages/nouns-contracts/script/DeployDAOV3NewContractsBase.s.sol
```diff
@@ -91,7 +91,7 @@ contract DeployDAOV3NewContractsBase is Script {
             FORK_DAO_PROPOSAL_THRESHOLD_BPS,
             FORK_DAO_QUORUM_VOTES_BPS
         );
-        daoV3Impl = new NounsDAOLogicV3();
+        daoV3Impl = new NounsDAOLogicV4();
```
davidbrai, line 94: V3 reference

packages/nouns-webapp/src/wrappers/nounsData.ts
```diff
@@ -93,8 +92,10 @@ const filterSigners = (
         ?.nounsRepresented.length || 0;
     // don't count votes from signers who have active or pending proposals
     // but include them in the list of signers to display with a note that they have an active proposal
-    const parentProposalIsUpdatable = proposalIdToUpdate && updatableProposalIds?.includes(proposalIdToUpdate ?? 0);
```
davidbrai, line 96: @eladmallel would be nice to avoid this formatting diff if possible

packages/nouns-contracts/script/DeployDAOV3NewContractsBase.s.sol
eladmallel: renamed

packages/nouns-webapp/src/wrappers/nounsData.ts
eladmallel: generally agree, but this is what is happening with our prettier config that we recently made sure was the same; had to change this file for a rename thing, and I have auto format on save. doesn't feel important enough to move slower and turn off auto formatting for these files.

wdyt?

packages/nouns-contracts/contracts/governance/NounsDAOInterfaces.sol
eladmallel: updated these names as well as the names of the DAO libraries

packages/nouns-contracts/contracts/libs/InMemoryMapping.sol
```diff
@@ -0,0 +1,84 @@
...
+ *********************************/
+
+pragma solidity ^0.8.19;
+
+/**
+ * @notice A library for short lived in memory mappings.
+ * This only works when the keys, in this case clientId, are of limited size so that they we can create
+ * an array in memory for all the possible values
+ */
+library InMemoryMapping {
```
eladmallel, line 23: the name of this lib gives the feeling that it's general purpose, when in fact the code is pretty specific for clientIds and balances, so I'm thinking giving it a more specific name is good.

wdyt? @davidbrai

packages/nouns-contracts/contracts/libs/InMemoryMapping.sol
```diff
@@ -0,0 +1,84 @@
...
+            m.indexes[clientId] = idx;
+            m.values[idx].clientId = clientId;
+        }
+        m.values[idx].balance = balance;
+    }
+
+    /**
+     * Increases the value of client `clientId` by `balance`
+     */
+    function inc(Mapping memory m, uint32 clientId, uint256 balance) internal pure {
```
eladmallel, line 66: feel like `balance` should have a more specific name, maybe `balanceDiff`, to make it super clear that it's a delta and not the value you're setting, wdyt?

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,591 @@
...
+    /**
+     * @notice Withdraws the balance of a client
+     * @dev The caller must be the owner of the NFT with id `clientId`
+     * @dev The maximum value of `amount` is one less than in `_clientBalances[clientId]`.
+     * This in order to leave 1 wei in storage and avoid expensive gas writes in future balance increases.
+     * @param clientId Which client balance to withdraw
+     * @param amount amount of withdraw
+     * @param to the address to withdraw to
+     */
+    function withdrawClientBalance(uint32 clientId, uint256 amount, address to) public whenNotPaused {
```
eladmallel, line 372: nit: I think it's nicer for `to` to come before `amount` to better mirror the ERC20 transfer interface

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,591 @@
...
+        return $.ethToken;
+    }
+
+    function admin() public view returns (address) {
+        RewardsStorage storage $ = _getRewardsStorage();
+        return $.admin;
+    }
+
+    /**
+     * @notice Returns the raw value from _clientBalances mapping. Usually you want to use `clientBalance`
```
eladmallel, line 499: worth expanding why we have this separation in this natspec, i.e. because of the 1 wei hack

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,591 @@
...
+    event ProposalRewardsUpdated(uint32 firstProposalId, uint32 lastProposalId);
+
+    /**
+     * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
+     *   CONSTANTS
+     * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
+     */
+
+    /// @notice The maximum priority fee used to cap gas refunds
+    uint256 public constant MAX_REFUND_PRIORITY_FEE = 2 gwei;
```
eladmallel, line 50: could be nice to extract gas refund logic to a library with internal functions, so these constants go in there and this contract becomes a bit easier to read. wdyt?

davidbrai: @eladmallel I actually started with doing a library because I wanted to reuse logic, but then didn't because this depended on the ERC20 token.
I can and pass the ERC20 token as a parameter.
won't be so reusable, but maybe easier to read

packages/nouns-contracts/test/foundry/DAOUpgrade/DAOUpgradeMainnetFork.t.sol
```diff
@@ -0,0 +1,185 @@
...
+        voteAndExecuteProposal(proposalId);
+
+        assertEq(expectedVotingDelay, NOUNS_DAO_PROXY_MAINNET.votingDelay());
+    }
+
+    function test_clientId_savedOnProposals() public {
+        uint32 expectedClientId = 42;
+        uint256 proposalId = propose(address(NOUNS_DAO_PROXY_MAINNET), 0, '', '', expectedClientId);
+
+        assertEq(expectedClientId, NOUNS_DAO_PROXY_MAINNET.proposalClientId(proposalId));
```
davidbrai, line 161: @eladmallel you're using `proposalClientId` and `proposalVoteClientData` below which I think are not actually used by the rewards contract any more.
I thought we could delete if they aren't used. But maybe it's useful to keep them? wdyt?

packages/nouns-contracts/test/foundry/DAOUpgrade/DAOUpgradeMainnetFork.t.sol
```diff
@@ -0,0 +1,185 @@
...
+
+import 'forge-std/Test.sol';
+import { Strings } from '@openzeppelin/contracts/utils/Strings.sol';
+import { NounsDAOLogicV4 } from '../../../contracts/governance/NounsDAOLogicV4.sol';
+import { ProposeDAOUpgradeMainnet } from '../../../script/DAOUpgrade/ProposeDAOUpgradeMainnet.s.sol';
+import { NounsToken } from '../../../contracts/NounsToken.sol';
+import { INounsDAOLogicV3 } from '../../../contracts/interfaces/INounsDAOLogicV3.sol';
+import { NounsDAOTypes } from '../../../contracts/governance/NounsDAOInterfaces.sol';
+
+abstract contract DAOUpgradeMainnetForkBaseTest is Test {
```
davidbrai, line 12: @eladmallel CI doesn't run fork tests at the moment.
I think we should make it run them, because we are now running without optimizer locally.
It will probably require setting up some alchemy/infura RPC into the github secrets. we may need soli's help for that

packages/nouns-contracts/test/foundry/DAOUpgrade/DAOUpgradeMainnetFork.t.sol
```diff
@@ -0,0 +1,185 @@
...
+
+        assertEq(prop.id, 493);
+        assertEq(prop.creationBlock, 19093670);
+        assertEq(prop.creationTimestamp, 0);
+
+        prop = NOUNS_DAO_PROXY_MAINNET.proposalsV3(474);
+
+        assertEq(prop.id, 474);
+        assertEq(prop.creationBlock, 18836862);
+        assertEq(prop.creationTimestamp, 0);
```
davidbrai, line 126: @eladmallel would be nice to also add a test that creationTimestamp is being written in new proposals

packages/nouns-webapp/src/wrappers/nounsData.ts
davidbrai: @eladmallel I would just do the edit in a text editor for the rename. but not a big deal either way

packages/nouns-contracts/contracts/libs/InMemoryMapping.sol
davidbrai: @eladmallel naming is hard. do you have an idea for a name?

packages/nouns-contracts/contracts/libs/InMemoryMapping.sol
davidbrai: sure, done

packages/nouns-contracts/contracts/Rewards.sol
```diff
@@ -0,0 +1,591 @@
...
+        for (uint256 i; i < settlements.length; ++i) {
+            INounsAuctionHouseV2.Settlement memory settlement = settlements[i];
+            if (settlement.clientId > 0) {
+                sawNonZeroClientId = true;
+                m.inc(settlement.clientId, settlement.amount);
+            }
+        }
+
+        uint16 auctionRewardBps = $.params.auctionRewardBps;
+        for (uint256 i = 1; i < m.nextAvailableIndex; ++i) {
```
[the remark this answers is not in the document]
davidbrai: @eladmallel made a refactor here: 900a7359db5dfbf0966c49965fbf26461c0cf5ff
wdyt?

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: a bit lazy to do the refactor right now, but noted!

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: done here: bbfe1ecee82bb3291a6c0102df92a86ab79330b6
wdyt?

packages/nouns-contracts/contracts/Rewards.sol
davidbrai: added

packages/nouns-contracts/contracts/libs/InMemoryMapping.sol
eladmallel: Maybe `ClientRewardsMemoryMapping`?

packages/nouns-contracts/contracts/Rewards.sol
eladmallel: @davidbrai I pushed this refactor :)

packages/nouns-contracts/test/foundry/DAOUpgrade/DAOUpgradeMainnetFork.t.sol
eladmallel: agreed and removed those functions

packages/nouns-contracts/test/foundry/DAOUpgrade/DAOUpgradeMainnetFork.t.sol
eladmallel: I agree it's good to have, let's ask for help!

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
```diff
@@ -0,0 +1,624 @@
...
+     * @param to the address to withdraw to
+     * @param amount amount of withdraw
+     */
+    function withdrawClientBalance(uint32 clientId, address to, uint256 amount) public whenNotPaused {
+        RewardsStorage storage $ = _getRewardsStorage();
+
+        require(ownerOf(clientId) == msg.sender, 'must be client NFT owner');
+        require(amount < $._clientBalances[clientId], 'amount too large');
+
+        $._clientBalances[clientId] -= amount;
```
eladmallel, line 407: @davidbrai I think we're reading the client's balance twice in this function and could instead cache it in a local var, wdyt?

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
```diff
@@ -0,0 +1,624 @@
...
+    function registerClient(
+        string calldata name,
+        string calldata description
+    ) public override whenNotPaused returns (uint32) {
+        RewardsStorage storage $ = _getRewardsStorage();
+
+        uint32 tokenId = super.registerClient(name, description);
+
+        // Increase the balance by one wei so that the slot is non zero when increased in the future
+        $._clientBalances[tokenId] += 1;
```
eladmallel, line 177: @davidbrai why is this `+=1` rather than `=1`?

davidbrai: @eladmallel because someone may pass a `clientId` before it was registered.
it's not a normal use case, but if it happens, I thought maybe better not to zero the balance
wdyt?

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
davidbrai: @eladmallel yes it can.
not touching it for now because maybe you'll change the way this works?

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
eladmallel: I think that makes sense, I think it can be helpful to add another inline comment explaining this, wdyt?

davidbrai: @eladmallel yes, but maybe not relevant if we're changing the way we track balances? then no need for the 1 wei hack

packages/nouns-contracts/contracts/libs/InMemoryMapping.sol
davidbrai: done

packages/nouns-contracts/contracts/governance/NounsDAOProxyV3.sol
davidbrai: fixed

packages/nouns-contracts/test/foundry/NounsDAOLogic/NounsDAOAdmin.t.sol
```diff
@@ -22,6 +22,121 @@ contract NounsDAOLogicAdminTest is NounsDAOLogicBaseTest {
 
     address[] tokens;
 
+    function test__setVotingDelay_onlyAdmin() public {
```
davidbrai, line 26: missing test for:
- `_setForkParams`
- `_withdraw`
- `_burnVetoPower`
- `_acceptVetoer`
- `_setPendingVetoer`
- `_setPendingAdmin` - check only admin can call
- `_acceptAdmin` - check only pending admin can call

eladmallel: * `_setForkParams`: I decided to skip because we have tests for all the individual params, but I gather that you think it's important so I don't mind adding anyway.
* `_withdraw`: tested in the file Withdraw.t.sol
* `_burnVetoPower`: tested in the file Veto.t.sol
* `_acceptVetoer`: tested in the file Veto.t.sol
* `_setPendingVetoer`: tested in the file Veto.t.sol
* `_setPendingAdmin`: will add onlyAdmin test
* `_acceptAdmin`: will add the only pending admin check

packages/nouns-contracts/script/Rewards/DeployRewardsMainnet.s.sol
```diff
@@ -0,0 +1,33 @@
...
+    address constant WETH_MAINNET = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+
+    function run() public returns (Rewards rewards) {
+        return
+            super.runInternal({
+                dao: INounsDAOLogic(DAO_PROXY_MAINNET),
+                auctionHouse: INounsAuctionHouseV2(AUCTION_HOUSE_MAINNET),
+                admin: VERBS_SAFE_MAINNET,
+                ethToken: WETH_MAINNET,
+                params: Rewards.RewardParams({
```
davidbrai, line 22: @eladmallel we need to discuss how to decide on these params.
most importantly the rewards bps.
but also to give extra thought to the other ones

packages/nouns-contracts/contracts/governance/NounsDAOLogicV4.sol
```diff
@@ -445,16 +515,45 @@ contract NounsDAOLogicV3 is NounsDAOStorageV3, NounsDAOEventsV3 {
      * @return A `ProposalCondensed` struct with the proposal data, not backwards compatible as it contains additional values
      * like `objectionPeriodEndBlock` and `signers`
      */
-    function proposalsV3(uint256 proposalId) external view returns (ProposalCondensed memory) {
+    function proposalsV3(uint256 proposalId) external view returns (ProposalCondensedV3 memory) {
         return ds.proposalsV3(proposalId);
     }
 
+    /**
+     * @notice Get a range of proposals, in the format of a samller struct tailored to client incentives rewards.
```
werme, line 523: "samller" -> "smaller"?

davidbrai: thanks, fixed

solimander (approved):

PR #832, opened 2024-01-30
fix: remove hardcoded reserve price from the webapp

0xMillz:


bagelface:
I think this is a great change that adds needed transparency on the frontend. I think this in combination with the non-optional setting would be an ideal solution. I'll make a PR from my branch into yours in case that's the route we all settle on.

0xMillz:
> I think this is a great change that adds needed transparency on the frontend. I think this in combination with the non-optional setting would be an ideal solution. I'll make a PR from my branch into yours in case that's the route we all settle on.

Glad you agree! I left it as a draft for right now because I haven't fully finished regression testing against my local and Sepolia networks. I'll probably have one more small change before marking it ready for review.

While I do have slight differences of opinion in the two PRs that focus on the contracts package, I felt they are better suited as review comments rather than bundling them into this PR.

PR #837, opened 2024-02-12
deploy script: use fee data to avoid gas errors

eladmallel:
did this fix after I got the error: `max fee per gas less than block base fee` when trying to deploy


eladmallel:
@davidbrai It's not certain that the old gas input doesn't work, it just fails under certain circumstances (the error message in the commit message), which could be avoided when inputting manual gas values.

I am also trying to spend minimal time on this now, so didn't want to get into changing the manual mode to receiving multiple inputs.

Since I basically don't use the manual mode ever, I could be ok with deleting that code and just always auto-deploying.

Shall we do that then?

davidbrai:
@eladmallel I think for this particular script it's fine, since it's only used for testnets

PR #839, opened 2024-04-05
Fix for rewards contract following audit

davidbrai:
This fix changes the way the proposal rewards calculate the auction revenue. Instead of taking the user provided lastProposalId's creation timestamp, we use the last *eligible* proposal that is <= lastProposalId. The reason is we don't want users creating bogus proposals just for controlling the end timestamp.

In addition, we also consider canceled proposals to be ineligible. This is to prevent a user with quorum votes to create proposals, vote, and cancel, just to get rewards.

### DIFF
--- a/packages/nouns-contracts/contracts/client-incentives/Rewards.sol
+++ b/packages/nouns-contracts/contracts/client-incentives/Rewards.sol
@@ -330,6 +330,7 @@ contract Rewards is
             votingClientIds
         );
         require(proposals.length > 0, 'at least one eligible proposal');
+        t.numEligibleProposals = proposals.length;
         $.nextProposalIdToReward = lastProposalId + 1;
 
         t.lastProposal = proposals[proposals.length - 1];
@@ -360,8 +361,6 @@ contract Rewards is
             uint256 votesInProposal = proposals[i].forVotes + proposals[i].againstVotes + proposals[i].abstainVotes;
             t.numEligibleVotes += votesInProposal;
         }
-
-        t.numEligibleProposals = proposals.length;
 
         //// Check that distribution is allowed:
         //// 1. One of the two conditions must be true:
--- a/packages/nouns-contracts/contracts/client-incentives/Rewards.sol
+++ b/packages/nouns-contracts/contracts/client-incentives/Rewards.sol
@@ -327,13 +327,14 @@ contract Rewards is
 
         uint16 proposalEligibilityQuorumBps_ = $.params.proposalEligibilityQuorumBps;
 
-        NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards(
-            t.nextProposalIdToReward,
-            lastProposalId,
-            proposalEligibilityQuorumBps_,
-            true,
-            votingClientIds
-        );
+        NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards({
+            firstProposalId: t.nextProposalIdToReward,
+            lastProposalId: lastProposalId,
+            proposalEligibilityQuorumBps: proposalEligibilityQuorumBps_,
+            excludeCanceled: true,
+            requireVotingEnded: true,
+            votingClientIds: votingClientIds
+        });
         require(proposals.length > 0, 'at least one eligible proposal');
         t.numEligibleProposals = proposals.length;
         $.nextProposalIdToReward = lastProposalId + 1;
@@ -353,16 +354,9 @@ contract Rewards is
         t.votingRewardForPeriod = (auctionRevenue * $.params.votingRewardBps) / 10_000;
 
         //// First loop over the proposals:
-        //// 1. Make sure all proposals have finished voting.
-        //// 2. Delete (zero out) proposals that are non elgibile (i.e. not enough For votes).
-        //// 3. Count the number of eligible proposals.
-        //// 4. Count the number of votes in eligible proposals.
+        //// 1. Count the number of votes in eligible proposals.
 
         for (uint256 i; i < proposals.length; ++i) {
-            // make sure proposal finished voting
-            uint endBlock = max(proposals[i].endBlock, proposals[i].objectionPeriodEndBlock);
-            require(block.number > endBlock, 'all proposals must be done with voting');
-
             uint256 votesInProposal = proposals[i].forVotes + proposals[i].againstVotes + proposals[i].abstainVotes;
             t.numEligibleVotes += votesInProposal;
         }
@@ -395,10 +389,9 @@ contract Rewards is
         );
 
         //// Second loop over the proposals:
-        //// 1. Skip proposals that were deleted for non eligibility.
-        //// 2. Reward proposal's clientId.
-        //// 3. Reward the clientIds that faciliated voting.
-        //// 4. Make sure all voting clientIds were included. This is meant to avoid griefing. Otherwises one could pass
+        //// 1. Reward proposal's clientId.
+        //// 2. Reward the clientIds that faciliated voting.
+        //// 3. Make sure all voting clientIds were included. This is meant to avoid griefing. Otherwises one could pass
         ////    a large array of votingClientIds, spend a lot of gas, and have that gas refunded.
 
         ClientRewardsMemoryMapping.Mapping memory m = ClientRewardsMemoryMapping.createMapping({
@@ -494,13 +487,14 @@ contract Rewards is
         for (uint32 i; i < numClientIds; ++i) {
             allClientIds[i] = i;
         }
-        NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards(
-            $.nextProposalIdToReward,
-            lastProposalId,
-            $.params.proposalEligibilityQuorumBps,
-            true,
-            allClientIds
-        );
+        NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards({
+            firstProposalId: $.nextProposalIdToReward,
+            lastProposalId: lastProposalId,
+            proposalEligibilityQuorumBps: $.params.proposalEligibilityQuorumBps,
+            excludeCanceled: true,
+            requireVotingEnded: true,
+            votingClientIds: allClientIds
+        });
 
         uint32[] memory sumVotes = new uint32[](numClientIds);
         for (uint256 i; i < proposals.length; ++i) {
@@ -692,10 +686,6 @@ contract Rewards is
         for (uint256 i = 0; i < s.length; ++i) {
             sum += s[i].amount;
         }
-    }
-
-    function max(uint256 a, uint256 b) internal pure returns (uint256) {
-        return a > b ? a : b;
     }
 
     /**
--- a/packages/nouns-contracts/contracts/client-incentives/Rewards.sol
+++ b/packages/nouns-contracts/contracts/client-incentives/Rewards.sol
@@ -294,14 +294,19 @@ contract Rewards is
 
     /**
      * @notice Distribute rewards for proposal creation and voting from the last update until `lastProposalId`.
-     * A proposal is eligible for rewards if for-votes/total-votes >= params.proposalEligibilityQuorumBps.
-     * Rewards are calculated by the auctions revenue during the period between the creation time of last proposal in
-     * the previous update until the current last proposal with id `lastProposalId`.
+     * A proposal is eligible for rewards if it wasn't canceled and for-votes/total-votes >= params.proposalEligibilityQuorumBps.
+     * Rewards are calculated by the auctions revenue during the period between the creation time of last processed
+     * eligible proposal in until the current last eligible proposal with id <= `lastProposalId`.
+     * One of two conditions must be true in order for rewards to be distributed:
+     * 1. There are at least `numProposalsEnoughForReward` proposals in this update
+     * 2. At least `minimumRewardPeriod` time has passed since the last update until the creation time of the last
+     *     eligible proposal in this update.
      * Gas spent is refunded in `ethToken`.
      * @param lastProposalId id of the last proposal to include in the rewards distribution. all proposals up to and
      * including this id must have ended voting.
      * @param votingClientIds array of sorted client ids that were used to vote on the eligible proposals in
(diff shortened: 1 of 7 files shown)


packages/nouns-contracts/contracts/client-incentives/Rewards.sol
```diff
@@ -320,11 +320,16 @@ contract Rewards is
         require(lastProposalId >= t.nextProposalIdToReward, 'bad lastProposalId');
         require(isSortedAndNoDuplicates(votingClientIds), 'must be sorted & unique');
 
+        uint16 proposalEligibilityQuorumBps_ = $.params.proposalEligibilityQuorumBps;
+
         NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards(
             t.nextProposalIdToReward,
             lastProposalId,
+            proposalEligibilityQuorumBps_,
```
eladmallel, line 328: I think the natspec of this function can add more clarity on the constraints we enforce, specific things on my mind:

* what makes a prop eligible: we are now filtering out canceled props
* define the minimum reward period and how it works, highlighting that we do not use block.timestamp
* votingClientIds: clarify that it reverts if they have zero votes on *eligible* props

packages/nouns-contracts/test/foundry/rewards/ProposalRewards.t.sol
```diff
@@ -294,6 +294,53 @@ contract ProposalRewardsTest is BaseProposalRewardsTest {
...
+        votingClientIds = [0];
+        rewards.updateRewardsForProposalWritingAndVoting({
+            lastProposalId: proposalId,
+            votingClientIds: votingClientIds
+        });
+
+        assertEq(rewards.clientBalance(clientId2), 0.15 ether); // 15 eth * 1%
+    }
+
+    function test_cantUseIneligibleProposalToPassTheMinimumPeriod() public {
```
eladmallel, line 322: maybe worth further explaining with comments the state you're creating here?

* we do not pass `numProposalsEnoughForReward`, hence it'll revert
* then we have the time filter which looks at the creation timestamp of the most recent eligible proposal - and that's the constraint an attacker might try to circumvent, and that's what we're showing won't work here

packages/nouns-contracts/test/foundry/rewards/ProposalRewards.t.sol
```diff
@@ -459,6 +506,21 @@ contract ProposalRewardsEligibilityTest is BaseProposalRewardsTest {
             votingClientIds: votingClientIds
         });
     }
+
+    function test_canceledProposalsAreIneligible() public {
+        params.proposalEligibilityQuorumBps = 7000;
```
eladmallel, line 511: I am confused as to why you're using 7000 here. looks like bidder has 8 votes and supply is 12 so bidder has 67% not 70%, meaning with this config they are below quorum, but I think you want them to have quorum and only be ineligible bc it was canceled no?

davidbrai: I copied the 7000 value from a test above. with 7000 it's still valid because of flooring:
floor(12*7000/10000) = 8

packages/nouns-contracts/test/foundry/rewards/ProposalRewards.t.sol
```diff
@@ -709,6 +771,51 @@ contract VotesRewardsTest is BaseProposalRewardsTest {
...
+        vote(bidder1, proposalId, 0, 'against', clientId1);
+        mineBlocks(VOTING_PERIOD);
+
+        uint32 proposalId2 = uint32(propose(bidder1, address(1), 1 ether, '', '', 'my proposal', 0));
+        mineBlocks(VOTING_DELAY + UPDATABLE_PERIOD_BLOCKS + 1);
+        vote(bidder1, proposalId2, 1, 'i support', clientId2);
+        mineBlocks(VOTING_PERIOD);
+        uint32[] memory votingClientIds = rewards.getVotingClientIds(proposalId2);
+
+        // doesn't revert
```
eladmallel, line 785: why would we expect it to revert?

e.g. if I'm thinking about the check that every client has non-zero votes, both clients 1 and 2 have votes here, so they would not fail that check.

can you specify what check would have reverted but does not thanks to the filtering working?

packages/nouns-contracts/contracts/governance/NounsDAOProposals.sol
```diff
@@ -731,6 +733,9 @@ library NounsDAOProposals {
         for (uint256 pid = firstProposalId; pid <= lastProposalId; ++pid) {
             proposal = ds._proposals[pid];
 
+            if (excludeCanceled && proposal.canceled) continue;
```
eladmallel, line 736: I don't think we have a test that runs through what happens when this exclude param is false?

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
```diff
@@ -354,26 +357,17 @@ contract Rewards is
...
-            }
-
-            // proposal is eligible for reward
-            ++t.numEligibleProposals;
-
             uint256 votesInProposal = proposals[i].forVotes + proposals[i].againstVotes + proposals[i].abstainVotes;
             t.numEligibleVotes += votesInProposal;
         }
 
+        t.numEligibleProposals = proposals.length;
```
eladmallel, line 364: nit: can we move this up, right after we get the proposals list?

packages/nouns-contracts/contracts/governance/NounsDAOProposals.sol
davidbrai: added a test: d003fe857d65dc76c8a60ef6c0d57ad4aae896f3

packages/nouns-contracts/contracts/governance/NounsDAOProposals.sol
```diff
@@ -731,6 +733,9 @@ library NounsDAOProposals {
...
+            if (proposal.forVotes < (proposal.totalSupply * proposalEligibilityQuorumBps) / 10_000) continue;
```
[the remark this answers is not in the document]
davidbrai: fixed: 515900ec83378bb8bb592f6d9b9d8e11a72b1bfb

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
davidbrai: 69bc31a158cc629dccdf27c54acf6ed69b4b5d22

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
davidbrai: updated natspec: 157d500e5257bcc22fe914f0ceef7e1fe93ea92f

packages/nouns-contracts/test/foundry/rewards/ProposalRewards.t.sol
davidbrai: added: 2669f0174a3524aabbed5da5d804534f1fe54de7

packages/nouns-contracts/test/foundry/rewards/ProposalRewards.t.sol
davidbrai: added comment: dbc71b8330b798079fa7b70d2f71a4c904135a00

packages/nouns-contracts/test/foundry/rewards/ProposalRewards.t.sol
davidbrai: added a comment to explain: 7b9e962b39815f8a6b96fc609c151749b2f2f5cb

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
```diff
@@ -320,11 +325,17 @@ contract Rewards is
...
         NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards(
             t.nextProposalIdToReward,
             lastProposalId,
+            proposalEligibilityQuorumBps_,
+            true,
             votingClientIds
         );
+        require(proposals.length > 0, 'at least one eligible proposal');
+        t.numEligibleProposals = proposals.length;
         $.nextProposalIdToReward = lastProposalId + 1;
```
dmitriia, line 339: Should be `$.nextProposalIdToReward = t.lastProposal + 1` (`t.lastProposal` setting needs to go up accordingly) otherwise there can be griefing.

davidbrai: You're right, documenting the potential griefing with more detail:
because the check that voting is over is done only on the filtered proposals returned from `proposalDataForRewards`, `lastProposalId` has only limited to be `<= nounsDAO.proposalCount()`.
Someone can provide a `lastProposalId` of a proposal that was just created and it would be processed as completed and rewards would be skipped for that proposal.

davidbrai: Thinking about it more, another potential issue:
The check that voting has finished is now done only on proposals returned from `proposalDataForRewards`.
Could there be a proposal earlier than `t.lastProposal` that has not yet finished voting, and has been filtered out in `proposalDataForRewards`?
If objection period is turned on for a proposal, then its voting period could end *after* a later proposal.
But, for objection period to be turn on it means that the proposal got at least enough votes to pass, using the DAO's dynamic quorum. So I think in practice there's low risk here.

Another potential situation in which an earlier proposal can finish voting after a later one is in case the DAO decreases the votingPeriod parameter.
In this case the code here could filter out the earlier proposal, even though it may get more votes in the future.

For these reasons, I think it's worthwhile to check that all proposals up to `nextProposalIdToReward` have finished voting.
With the current code design, this means adding this check to the loop in `proposalDataForRewards`:
```
            // make sure proposal finished voting
            uint endBlock = max(proposals[i].endBlock, proposals[i].objectionPeriodEndBlock);
            require(block.number > endBlock, 'all proposals must be done with voting');
```

@eladmallel @dmitriia please lmk what you think

dmitriia: > You're right, documenting the potential griefing with more detail: because the check that voting is over is done only on the filtered proposals returned from `proposalDataForRewards`, `lastProposalId` has only limited to be `<= nounsDAO.proposalCount()`. Someone can provide a `lastProposalId` of a proposal that was just created and it would be processed as completed and rewards would be skipped for that proposal.

I think this is correct, any proposal in a early stage, having votes below the threshold, can be attacked this way. Also, it can be an indirect/by-product impact with other proposal's beneficiaries just wanting to receive their rewards, not intending to grief the unfinished ones in the first place.

dmitriia: Since only `against` votes can be casted during the objection period (so it cannot be not eligible before and eligible after just by `forVotes` accumulation), the only viable scenarios seems to be `votingPeriod` or `proposalEligibilityQuorumBps` decreases. If `votingPeriod` was decreased between two proposals then a proposal that still can go beyond the eligibility threshold can be filtered out if the next one is already concluded. If `proposalEligibilityQuorumBps` was decreased then running the rewards allocation just before that decrease can deprive a proposal based on the old rules. This is, however, a kind of usual race condition on configuration changes.

> For these reasons, I think it's worthwhile to check that all proposals up to `nextProposalIdToReward` have finished voting. With the current code design, this means adding this check to the loop in `proposalDataForRewards`:
> 
> ```
>             // make sure proposal finished voting
>             uint endBlock = max(proposals[i].endBlock, proposals[i].objectionPeriodEndBlock);
>             require(block.number > endBlock, 'all proposals must be done with voting');
> ```
> 
> @eladmallel @dmitriia please lmk what you think

I support this. I think it has to go after `cancel` check as `Canceled` is a final state and it will not matter how far is `endBlock`.

Also, it will cover the `votingPeriod` change, but not `proposalEligibilityQuorumBps` change. It looks to be fine as the latter is a kind of expected behavior.

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
```diff
@@ -320,11 +325,18 @@ contract Rewards is
...
+        NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards({
+            firstProposalId: t.nextProposalIdToReward,
+            lastProposalId: lastProposalId,
+            proposalEligibilityQuorumBps: proposalEligibilityQuorumBps_,
+            excludeCanceled: true,
+            requireVotingEnded: true,
+            votingClientIds: votingClientIds
+        });
+        require(proposals.length > 0, 'at least one eligible proposal');
+        t.numEligibleProposals = proposals.length;
```
eladmallel, line 339: perhaps it's best now to remove `t.numEligibleProposals` and instead just use `proposals.length`?
it feels confusing to have this extra var, it makes me assume there's a difference between the proposals list and which are considered eligible, when in fact now they are all eligible.

wdyt?

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
```diff
@@ -320,11 +325,18 @@ contract Rewards is
         require(lastProposalId >= t.nextProposalIdToReward, 'bad lastProposalId');
         require(isSortedAndNoDuplicates(votingClientIds), 'must be sorted & unique');
 
-        NounsDAOTypes.ProposalForRewards[] memory proposals = nounsDAO.proposalDataForRewards(
-            t.nextProposalIdToReward,
-            lastProposalId,
-            votingClientIds
-        );
+        uint16 proposalEligibilityQuorumBps_ = $.params.proposalEligibilityQuorumBps;
```
eladmallel, line 328: nit: do we still need to cache this variable? I think we're only using it once now?

davidbrai: fixed 1ebd2c4666716613632fc1a104ebfbbbbafdbe05

packages/nouns-contracts/contracts/client-incentives/Rewards.sol
davidbrai: fixed 1ebd2c4666716613632fc1a104ebfbbbbafdbe05

dmitriia:
Looks ok for me now

PR #840, opened 2024-04-17
DAO accounting for the nouns fungible token

eladmallel:
about potentially excluding Nouns that are inside the $nouns contract from governance


packages/nouns-contracts/contracts/governance/NounsDAOAdmin.sol
```diff
@@ -537,12 +546,10 @@ library NounsDAOAdmin {
...
     /**
-     * @notice Admin function for zeroing out the state variable `voteSnapshotBlockSwitchProposalId`
-     * @dev We want to zero-out this state slot so we can remove this temporary variable from contract code and 
-     * be ready to reuse this slot.
+     * @notice Admin function for setting the canonical Nouns Fungible Token address
      */
-    function _zeroOutVoteSnapshotBlockSwitchProposalId() external onlyAdmin {
-        ds().voteSnapshotBlockSwitchProposalId = 0;
+    function _setNounsFungibleToken(address newNounsFungibleToken) public onlyAdmin {
+        ds().nounsFungibleToken = newNounsFungibleToken;
```
davidbrai, line 552: add event?

packages/nouns-contracts/contracts/governance/NounsDAOLogicV4.sol
```diff
@@ -925,9 +925,5 @@ contract NounsDAOLogicV4 is NounsDAOStorage, NounsDAOEventsV3 {
         return address(ds.timelockV1);
     }
```
davidbrai, line 927: maybe add a view function for `nounsFungibleToken`?

packages/nouns-contracts/contracts/governance/NounsDAOAdmin.sol
eladmallel: I've been thinking we don't really use these events ever; esp this param will be set once and probably never again, unlike other gov params. can you please share your thoughts on why add an event here?

davidbrai: the general rule of thumb I use is that whenever storage is modified, an event should be fired.
I think there should be a stronger reason for not-adding than for adding

eladmallel: added an event

packages/nouns-contracts/contracts/governance/NounsDAOAdmin.sol
```diff
@@ -528,7 +541,11 @@ library NounsDAOAdmin {
      * @param timelockV1 the new timelockV1 contract
      * @param admin the new admin address
      */
-    function _setTimelocksAndAdmin(address timelock, address timelockV1, address admin) external onlyAdmin {
```
davidbrai, line 531: @eladmallel looks like your formatter is misconfigured.
this line is < 120 chars

packages/nouns-contracts/contracts/governance/fork/NounsDAOFork.sol
```diff
@@ -216,11 +219,20 @@ library NounsDAOFork {
 
     /**
      * @notice Returns the number of nouns in supply minus nouns owned by the DAO, i.e. held in the treasury or in an
-     * escrow after it has closed.
+     * escrow after it has closed, minus nouns hold via holding the nouns fungible token.
      * This is used when calculating proposal threshold, quorum, fork threshold & treasury split.
      */
     function adjustedTotalSupply(NounsDAOTypes.Storage storage ds) internal view returns (uint256) {
```
davidbrai, line 225: @eladmallel something to consider and maybe check:
we're increasing the gas cost of this function.
by how much?
which functions call this and get their gas cost increased due to it?

eladmallel: the change costs 2 cold storage reads (2100 gas x 2 = 4200) and an external call from the DAO to the fungible token.

`adjustedTotalSupply` is used in:

* submitting proposals
* fork functions like execute fork and join fork

this change adds less than 1% to the gas cost of proposing, therefore I think it's definitely worth it.

volkyeth:
Closing this since it won't be integrated into the DAO

PR #864, opened 2024-10-12
NounsDescriptorV3

EricDevito:
NounsDescriptorV3 upgrade. Includes contracts, tests, and hardhat tasks.


packages/nouns-contracts/contracts/NounsArt.sol
```diff
@@ -327,6 +327,154 @@ contract NounsArt is INounsArt {
         emit GlassesAdded(imageCount);
     }
 
+    /**
+     * @notice Replace current batch of body images with new ones.
```
eladmallel, line 331: @EricDevito I believe this is inaccurate and worth fixing in all natspecs. In other functions like `addHeads` or `addHeadsFromPointer` we say `Add a batch` in the natspec, because you are indeed adding a new batch into the Trait's dataset, which consists of multiple batches or pages.

In your code you are replacing the entirety of the trait data with this new data. This is an important distinction worth emphasizing.

wdyt?

packages/nouns-contracts/contracts/NounsDescriptorV3.sol
```diff
@@ -0,0 +1,618 @@
...
+
+    /**
+     * @notice Get the color palette pointer for the passed part.
+     */
+    function _getPalette(bytes memory part) private view returns (bytes memory) {
+        return art.palettes(uint8(part[0]));
+    }
+
+    /**
+     * @notice Replace current batch of accessories images with new ones.
```
eladmallel, line 472: @EricDevito similar to my comment in the art file, seems important to clarify that we're replacing the entire trait, not just a "current batch".

packages/nouns-contracts/tasks/index.ts
```diff
@@ -2,11 +2,10 @@ export * from './accounts';
 export * from './create-proposal';
 export * from './populate-descriptor';
 export * from './descriptor-art-to-console';
-export * from './deploy-descriptor-v2';
+export * from './deploy-descriptor-v3';
 export * from './populate-descriptor-via-proposal';
 export * from './deploy-test-token';
 export * from './upgrade-descriptor-via-proposal';
-export * from './deploy-local-dao-v3';
```
eladmallel, line 9: @EricDevito why remove this script? seems unrelated

packages/nouns-contracts/tasks/populate-descriptor-v3.ts
```diff
@@ -0,0 +1,65 @@
+import { task, types } from 'hardhat/config';
+import ImageData from '../files/image-data-v2.json';
+import { dataToDescriptorInput } from './utils';
+
+task('populate-descriptor-v3.', 'Populates the descriptor with color palettes and Noun parts')
```
eladmallel, line 5: @EricDevito why do we need this script when we have the original `populate-descriptor` script that you've also changed to use V3 and seems to do the same things?

packages/nouns-contracts/test/descriptorV3.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    const itemCount = ProposalImages[0].glasses?.itemCount ?? 0;
+
+    const glassesCountBefore = await nounsDescriptor.glassesCount();
+
+    console.log(`glassesCountBefore: ${glassesCountBefore}`);
+    await nounsDescriptor.updateGlasses(newEncodedCompressedTrait, decompressedLength, itemCount, {
+      gasLimit: 30000000,
+    });
+
+    const glassesCountAfter = await nounsDescriptor.glassesCount();
```
eladmallel, line 37: @EricDevito I believe the descriptor code insists that the image count before and after be the same? -- if so, why not asset this equality here in the test?

packages/nouns-contracts/test/descriptorV3.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    const glassesAtIdex = await nounsDescriptor.glasses(glassesTraitIndex);
```
eladmallel, line 38: @EricDevito I believe there's a type here, should be `Index`

packages/nouns-contracts/test/descriptorV3.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import ImageData from '../files/image-data-v2.json';
+import ProposalImages from '../files/proposal-test-traits.json';
+import { LongestPart } from './types';
+import { deployNounsDescriptorV3, populateDescriptorV2 } from './utils';
+import { ethers } from 'hardhat';
+
+chai.use(solidity);
+const { expect } = chai;
+
+const eyesred =
```
eladmallel, line 13: @EricDevito  what is this used for?

packages/nouns-contracts/test/descriptorV3.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  afterEach(async () => {
+    await ethers.provider.send('evm_revert', [snapshotId]);
+  });
+
+  it('updates all glasses with updated trait at index 0', async () => {
+    return await checkGlassesAfterUpdate();
+  });
+
+
+  // Unskip this test to validate the encoding of all parts. It ensures that no parts revert when building the token URI.
```
eladmallel, line 100: @EricDevito you probably want to clean up this commented-out section?

packages/nouns-contracts/test/descriptorV3.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  beforeEach(async () => {
+    snapshotId = await ethers.provider.send('evm_snapshot', []);
+  });
+
+  afterEach(async () => {
+    await ethers.provider.send('evm_revert', [snapshotId]);
+  });
+
+  it('updates all glasses with updated trait at index 0', async () => {
```
eladmallel, line 95: @EricDevito the description of the test isn't very clear IMO. can you please consider updating it such that it clarifies what you are testing exactly. for example, are you replacing the art of the first glasses item from X to Y?

eladmallel: And how do you know that the update didn't mess up other glasses?

packages/nouns-contracts/test/foundry/helpers/DeployUtils.sol
```diff
@@ -90,9 +91,24 @@ abstract contract DeployUtils is Test, DescriptorHelpers {
         return descriptorV2;
     }
 
+    function _deployAndPopulateV3() internal returns (NounsDescriptorV3) {
```
eladmallel, line 94: @EricDevito I noticed there are a bunch of descriptor-related tests in solidity (foundry), and not all of them are here in your diff. The main file is `NounsDescriptorV2.t.sol`. Seems like it has relevant tests you want to make sure apply to V3 as well?

packages/nouns-contracts/test/foundry/NounsDescriptorV3.t.sol
```diff
@@ -0,0 +1,742 @@
...
+        );
+        assertTrue(
+            imageSlice.endsWith(
+                '<rect width="60" height="10" x="100" y="160" fill="#ff638d" /><rect width="60" height="10" x="170" y="160" fill="#ff638d" /></svg>'
+                    .toSlice()
+            )
+        );
+    }
+
+    function testUpdateGlasssWorks() public {
```
eladmallel, line 711: @EricDevito I'm wondering now that you added this test code, why not include in it the same checks you have in the typescript test file, and then just remove the typescript test?

we are generally trying to be foundry maxis as much as we can.

wdyt?

davidbrai (approved):

PR #869, opened 2024-11-20
DRAFT: stream escrow subgraph

davidbrai:


packages/nouns-subgraph/src/stream-escrow.ts
```diff
@@ -0,0 +1,117 @@
...
+  const s = getStreamEscrowState();
+  s.totalAmountStreamedToDAO = s.totalAmountStreamedToDAO.plus(event.params.amount);
+  s.save();
+}
+
+export function handleStreamCreated(event: StreamCreated): void {
+  const nounId = event.params.nounId.toString();
+  const streamId = genericUniqueId(event);
+
+  const s = new Stream(streamId);
```
davidbrai, line 38: @eladmallel perhaps we can also save the currentTick here as the startTick?

packages/nouns-subgraph/src/stream-escrow.ts
```diff
@@ -0,0 +1,117 @@
...
+  son.save();
+}
+
+export function handleStreamFastForwarded(event: StreamFastForwarded): void {
+  const nounId = event.params.nounId.toString();
+
+  const ff = new StreamFastforward(genericUniqueId(event));
+  ff.createdTimestamp = event.block.timestamp;
+  ff.createdBlock = event.block.number;
+  ff.stream = nounId;
```
davidbrai, line 64: @eladmallel this looks wrong. should be `getOrCreateStreamsOfNoun(nounId).currentStream` I think?

packages/nouns-subgraph/src/stream-escrow.ts
```diff
@@ -0,0 +1,117 @@
...
+  const s = getCurrentStream(nounId);
+  s.lastTick = event.params.newLastTick;
+  s.streamLengthInTicks = s.streamLengthInTicks - event.params.ticksToForward.toI32();
+  s.save();
+}
+
+export function handleStreamCanceled(event: StreamCanceled): void {
+  const nounId = event.params.nounId.toString();
+
+  const s = getCurrentStream(nounId);
```
davidbrai, line 78: @eladmallel I think we should update `ethStreamedPerTick` here. Maybe we should add it to the StreamCanceled event. wdyt?

davidbrai: and maybe also in `StreamFastForwarded`: if the stream was fast forwarded all the way, then the stream finishes and `ethStreamedPerTick` decreases. so maybe we should add it there also

packages/nouns-subgraph/src/stream-escrow.ts
eladmallel: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/869/commits/47394bdf71a82e2d4136e0be189eccc2bb795d08

packages/nouns-subgraph/src/stream-escrow.ts
eladmallel: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/869/commits/8395db0105202ca5ae175c1de8722eb8cbbef9da

packages/nouns-subgraph/src/stream-escrow.ts
eladmallel: fixed: https://github.com/nounsDAO/nouns-monorepo/pull/869/commits/cf241d1bc76e344ca7a4d46cdd4a42ededb7a415

PR #871, opened 2024-12-03
DRAFT: fix audit issue 184

davidbrai:


packages/nouns-contracts/contracts/StreamEscrow.sol
```diff
@@ -75,6 +75,7 @@ contract StreamEscrow is IStreamEscrow {
     ) {
         daoExecutor = daoExecutor_;
         ethRecipient = ethRecipient_;
+        require(nounsRecipient_ != address(0), 'zero address');
```
eladmallel, line 78: can you please put this requirement at the top of the function?

packages/nouns-contracts/contracts/StreamEscrow.sol
```diff
@@ -279,6 +280,7 @@ contract StreamEscrow is IStreamEscrow {
      * @notice Allows the DAO to set the address that the Nouns tokens will be sent to when streams are canceled.
      */
     function setNounsRecipient(address newAddress) external onlyDAO {
+        require(newAddress != address(0), 'zero address');
```
eladmallel, line 283: can you please add natspec that explains why we aren't accepting zero address, i.e. explain the potential griefing scenario briefly? something like:

"newAddress cannot be the zero address because that would cause all stream cancellations to revert, since the Nouns token contract does not permit sending tokens to the zero address"

davidbrai: fixed

packages/nouns-contracts/contracts/StreamEscrow.sol
davidbrai: fixed

eladmallel (approved):

PR #876, opened 2024-12-19
Auction bidder sanction check

eladmallel:


dmitriia (commented):
It looks clean implementation wise. Some thoughts on the related risks and design implications:

1. `sanctionsOracle` becomes an another point of trust: it can turn rogue, block all other bids by marking the bidders sanctioned (say just when the `isSanctioned()` request comes from the Auction House, behaving normally otherwise) and keep winning the auctions for non-material amounts until auctions become paused and `sanctionsOracle` gets replaced

2. There is also a possibility of collision by an attacker and `sanctionsOracle` controlling party, which can selectively revert bids when the attacker pursues the auctioned noun, splitting the difference between realized and current market prices

3. If there be a known algorithm (or even a heuristics being correct frequently enough) for marking an address sanctioned, say a tx from an already sanctioned address with a value above some threshold makes its recipient sanctioned too, then it can be exploited by any third party that controls the sanctioned address, i.e. they can bid small on each auction, then front run all the other bids with such tx from the sanctioned address they control, reverting all the other bids and getting the noun at a depressed price

4. In order to control for operational mistakes it can be recommended to check Oracle availability, e.g. `require(!sanctionsOracle_.isSanctioned(address(this)), 'Sanctions Oracle malfunction')` on setting the new Oracle in `initialize()` and `setSanctionsOracle()`

5. `sanctionsOracle` non-malicious malfunction will freeze auctions: `createBid()` cannot proceed when `sanctionsOracle_.isSanctioned(account)` call reverts. If the sale without sanctions check is feasible consider wrapping the call in a try-catch block and place Oracle unavailability case logic in the catch section. This paired with (6) can suggest some auction time extension logic (i.e. Oracle was checked at the start, as (6) suggests, then it became unavailable, so the auction was unavailable too for some time in the middle and should be extended to a degree)

6. In order to reduce the number of situations when Oracle was unavailable for a part of the particular auction timeline, so the amount of bids and the resulting price was lower due to that only, consider checking the Oracle availability in `_createAuction()` (reverting if Oracle isn't available) and `_settleAuction()` (if (5) won't be implemented, so `createBid()` fails silently and don't affect the auction, there is a lighter alternative: not settling, but extending the auction if Oracle isn't available on `_settleAuction()`)

davidbrai (approved):

PR #881, opened 2025-02-04
hotfix for rendering nouns with traits missing from nouns/assets

EricDevito:
call descriptor contract to generate noun image when seed parts of a given noun are missing from nouns/assets


lastpunk9999:
Thanks for this. prefer to use `@nomonouns/assets` package as a 1:1 drop in

PR #884, opened 2025-02-17
Switch to @noundry/nouns-assets

volkyeth:
nouns.wtf is out of date again, missing the [Lilac body](https://www.nouns.camp/proposals/741)

This PR switches from the @nomonouns/assets package to @noundry/nouns-assets
that is a slightly modified version of @nouns/assets which [uses more up-to-date assets if available](https://github.com/volkyeth/noundry/blob/26f4307fb9743c329771dc2a755e06b1857c82ae/packages/nouns-assets/src/image-data.ts#L12-L14), and fallbacks to the package assets.

It also adds a script fetching the latest assets from https://assets.noundry.wtf/nouns/image-data.js
this script injects the most up-to-date assets for @noundry/nouns-assets to use.

Those assets are generated directly from the onchain data, so they'll always be up-to-date. No more need to update the package and push a new deployment all the time.

As a proof of concept, the current image-data.json from @noundry/nouns-assets [does not contain the Lilac body](https://github.com/volkyeth/noundry/blob/26f4307fb9743c329771dc2a755e06b1857c82ae/packages/nouns-assets/src/image-data.json#L362-L366). It is nevertheless loaded from the assets.noundry.wtf API and is available on the app:

CleanShot 2025-02-17 at 14 50 37@2x
(tested locally)

Obs: although this solution adds a dependency to the assets.noundry.wtf API, in a scenario where it's unavailable, the app will still load as usual, falling back to the assets included with @noundry/nouns-assets

### DIFF
--- a/packages/nouns-webapp/package.json
+++ b/packages/nouns-webapp/package.json
@@ -15,7 +15,7 @@
     "@lingui/macro": "3.13.3",
     "@lingui/react": "3.13.3",
     "@netlify/functions": "^0.7.2",
-    "@nomonouns/assets": "0.11.0",
+    "@noundry/nouns-assets": "^1.0.0",
     "@nouns/sdk": "^0.4.0",
     "@reduxjs/toolkit": "^1.6.0",
     "@testing-library/jest-dom": "^5.11.4",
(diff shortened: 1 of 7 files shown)


volkyeth:
Will actually integrate the auto artwork updates into @nouns/assets and switch back to that one.

Figured out a better way of doing it by leveraging https://www.jsdelivr.com/ instead of relying directly on https://assets.noundry.wtf/nouns/image-data.js, which will def offer superior speed and uptime

PR #886, opened 2025-03-03
Fix: missing command `run-local`

brunoamorimdev:
- Updated the deploy-local-dao-v3.ts script with valid values for executing it
- Updated the `run-local` command using a valid hardhat custom task


volkyeth:
done in #902

Issue #895, opened 2025-04-26
migrate from ethers to wagmi, viem, [appkit](https://github.com/reown-com/appkit)


volkyeth:
while I never used [appkit](https://github.com/reown-com/appkit) myself, it seems to be the successor to the deprecated web3modal and to be well regarded by devs in the space

could also consider the simpler [connectkit](https://github.com/family/connectkit) if that one's too much. I've used it on a bunch of projects and it's a solid pick

volkyeth:
wagmi migration is done. connectkit or appkit can be considered later

Issue #894, opened 2025-04-26
switch from lerna to pnpm workspaces


volkyeth:

nekofar:
taking this one, gonna work on migrating from lerna and yarn to pnpm.

PR #902, opened 2025-04-28
Feature/modernize stack

### LINKED ISSUE #897: update node 16 -> 22

volkyeth:
Closes #897


volkyeth:
@nekofar Netlify has node locked at 16
Will try to expedite the migration to providers under our control

Forge tests green in here

The problem was most likely access to the MAINNET_RPC secret on your fork. Weird, seems to be undocumented behavior.

nekofar:
@volkyeth ok good, gonna spend some time on #894 migrating from `lerna` and `yarn` to `pnpm`.

volkyeth:
nvm, I was able to override node version in 25cd729

now build has some more issues to fix

nekofar:
> nvm, I was able to override node version in [25cd729](https://github.com/nounsDAO/nouns-monorepo/commit/25cd7296a8311822972c1459b3019901cfe87baa)
> 
> now build has some more issues to fix

hey, i think `@babel/plugin-syntax-dynamic-import` is a peerDependency and for some reason it's not installed in Netlify environment. probably just adding it directly to `nouns-webapp` should solve it.

nekofar:
working with `lingui` is not easy, it can break code real quick, and from my experience, it even happens on minor upgrades. i wouldn’t mess with it personally; would suggest handling it as a separate issue. 😁

https://github.com/nounsDAO/nouns-monorepo/pull/902/commits/8b2f1d4d331f6186f67ef5da7b16c4c286be2564

nekofar:
> > nvm, I was able to override node version in [25cd729](https://github.com/nounsDAO/nouns-monorepo/commit/25cd7296a8311822972c1459b3019901cfe87baa)
> > now build has some more issues to fix
> 
> hey, i think `@babel/plugin-syntax-dynamic-import` is a peerDependency and for some reason it's not installed in Netlify environment. probably just adding it directly to `nouns-webapp` should solve it.

hey, can you change build setting to this? I think the issue with current build is Netlify just get `nouns-webapp` folder for build not whole project:

netlify-build

volkyeth:
> hey, i think `@babel/plugin-syntax-dynamic-import` is a peerDependency and for some reason it's not installed in Netlify environment. probably just adding it directly to `nouns-webapp` should solve it.

yeah, there's a whole bunch of `@babel/*` packages in that situation. tried adding them all as dependencies, but then I get some ESM problems. will try pinpointing all the right versions


> working with `lingui` is not easy, it can break code real quick, and from my experience, it even happens on minor upgrades. i wouldn’t mess with it personally; would suggest handling it as a separate issue. 😁

it seems to be broken already. running `i18n:extract` seems to comment out every translation string, so it's not picking them up from the code.

probably best not to mess with it rn and take care of that after we're done revamping the tooling

nekofar:
i've deployed the monorepo to netlify, and it built properly. auction isn't working though, which could be because i didn't set all env vars; just the ones i knew from the `.env` file. i think it's hard to determine the issue here when we don't know the current exact configuration of netlify.



https://680fca3bbc16f600644ed4e7--extraordinary-kelpie-7d288c.netlify.app/

volkyeth:
> hey, can you change build setting to this? I think the issue with current build is Netlify just get `nouns-webapp` folder for build not whole project:

looks like [netlify.toml has precedence](https://docs.netlify.com/configure-builds/file-based-configuration/), so I think we can. will try and push those

> i've deployed the monorepo to netlify, and it built properly. auction isn't working though, which could be because i didn't set all env vars; just the ones i knew from the .env file. i think it's hard to determine the issue here when we don't know the current exact configuration of netlify.

I got it working locally. You're probably missing `REACT_APP_INFURA_PROJECT_ID` which is mandatory even when setting a `REACT_APP_MAINNET_JSONRPC` (don't ask me why)

volkyeth:
@nekofar still same issue on latest build. also getting the same error locally

maybe move on with the pnpm migration and try enabling `auto-install-peers`?

nekofar:
> @nekofar still same issue on latest build. also getting the same error locally
> 
> maybe move on with the pnpm migration and try enabling `auto-install-peers`?

i'll work on `pnpm` but meanwhile it seems to be everything ok here:
https://680fd4164f54411bbd85cbdb--extraordinary-kelpie-7d288c.netlify.app/

alough i've deployed up until https://github.com/nounsDAO/nouns-monorepo/pull/902/commits/6243bba9b855aeb8e770fc7e764533e153d48769

volkyeth:
ok, got access to Netlify and fixed the settings to match the ones above.

deployment is working now:
https://6810e4df263ec5a230583607--nouns-home.netlify.app/

will double-check everything to see if we're good, but already found some issues:
CleanShot 2025-04-29 at 12 07 01

volkyeth:
Treasury issue already happens in prod. most likely ethers shenanigans
we can fix later in #895

PR #950, opened 2025-05-03
feat: migrate `nouns-webapp` from `ethers` to `wagmi` stack

nekofar:
It is related to issue #895.

* removed dead deps (make-plural, history, web3-react, @usedapp/core, ethers, react-scripts, @davatar/react, bignumber.js, etc)
* bumped dev-deps across eslint, prettier, eslint-plugin-turbo, vitest, mocha/chai, solidity-coverage, netlify-cli, ts-node removal, husky + lint-staged tweaks
* replaced Web3ReactProvider and usedapp hooks with wagmi + viem for useBlockNumber, useAccount, usePublicClient, signature/timestamp handling
* unified numeric types: BigInt for block numbers, Number for vote counts; added null-checks everywhere
* simplified hook APIs and args and optimized imports with alias paths
* overhauled proposal/candidate parsing: added strict GraphQL types, safer parsing logic, improved candidate version handling, null/undefined safeguards
* updated subgraph data parsing, improved forks data and escrow event hooks
* refactored proposal creation, processing, vote state, feedback handling, quorum logic, cancelCandidate args, and UI rendering logic
* enhanced UI components: ENS avatars in wallet dropdown, ShortAddress updates, LinkIcon/NogglesLogo swap, improved ExplorePage and Auction nav components
* added tests: ABIUpload, AuctionNavigation, BidHistoryItem
* style tweaks: removed semicolons, flattened ESLint config, prettier-ignore updates
* GraphQL tooling: added codegen config, custom scalars, prebuild scripts, GraphQL schema types for subgraphs and proposal candidates


nekofar:
@volkyeth this one is almost ready, but changes from `master` on https://github.com/nounsDAO/nouns-monorepo/commit/ed320235529eb356c9f7fc920cea70063e2e56ee is to complex after 290 commits to get resolved. I'll look into it later.

nekofar:
@volkyeth Just ran through all the processes and everything looks solid, ping me if I missed anything.

volkyeth:
CleanShot 2025-05-16 at 14 54 02@2x

Switch wallet and Disconnect buttons are not working

packages/nouns-webapp/wagmi.config.ts
```diff
@@ -0,0 +1,132 @@
+import { defineConfig } from '@wagmi/cli';
```
volkyeth, line 1: wondering if this shouldn't go on @nouns/sdk instead. it has value as a standalone package

nekofar: might work but it breaks the current sdk. we need a roadmap for @nouns/assets and @nouns/sdk and that’s outside this PR. i do agree some of this code should live in reusable packages.

nekofar:
> @nekofar CleanShot 2025-05-16 at 14 54 02@2x
> 
> Switch wallet and Disconnect buttons are not working
> 
> edit
> 
> got it working after clearing site data. we should probably pinpoint what's causing that and ensure whoever opens up the app doesn't get stuck without being able to connect.
> 
> maybe clear all the domain local storage on first time opening it, and then set some key to prevent it from clearing on next visits?



> @nekofar CleanShot 2025-05-16 at 14 54 02@2x
> 
> Switch wallet and Disconnect buttons are not working
> 
> edit
> 
> got it working after clearing site data. we should probably pinpoint what's causing that and ensure whoever opens up the app doesn't get stuck without being able to connect.
> 
> maybe clear all the domain local storage on first time opening it, and then set some key to prevent it from clearing on next visits?



> @nekofar CleanShot 2025-05-16 at 14 54 02@2x
> 
> Switch wallet and Disconnect buttons are not working
> 
> edit
> 
> got it working after clearing site data. we should probably pinpoint what's causing that and ensure whoever opens up the app doesn't get stuck without being able to connect.
> 
> maybe clear all the domain local storage on first time opening it, and then set some key to prevent it from clearing on next visits?

We alerdy have a `Clear WalletConnect Data` link under `WalletConnectModal`, but It could be done otherways as well. 🤔

eslint.config.mjs
volkyeth, line 1: don't think running eslint with a global config is a good idea.
I had problems with it while working on #970 because ponder has its own eslint plugin which seems to register custom aliases, and it was failing with the global eslint config.

why should we have package-specific plugins on the global config? e.g. lingui

I think it's best to leave each package with its own esling config, and if we must keep a top level one, stick only to the settings that make sense for all the packages and extend it from the individual packages eslint config.

nekofar: > I think it's best to leave each package with its own esling config, and if we must keep a top level one, stick only to the settings that make sense for all the packages and extend it from the individual packages eslint config.

With ESLint v9 you should just use one config; it cuts down on conflicts. Define scopes for each plugin and rule right there in that file. Might need to tweak it for new scopes though 🤷‍♂️

packages/nouns-webapp/src/components/AddNounsToForkModal/index.tsx
```diff
@@ -105,23 +110,40 @@ export default function AddNounsToForkModal(props: Props) {
       addNounsToEscrow(selectedNouns);
     } else {
       setIsTwoStepProcess(true);
-      setApproval(config.addresses.nounsDAOProxy, true);
+      if (setApproval) {
```
volkyeth, line 113: this one is not working for some reason

volkyeth:
when navigating directly to a Noun page, app breaks. e.g. http://localhost:3000/noun/1508

```
Uncaught TypeError: Cannot convert a BigInt value to a number
    at Array.sort (<anonymous>)
    at useNounActivity (nounActivity.ts:294:6)
    at ProfileActivityFeed (index.tsx:39:36)
    ...
```

packages/nouns-webapp/src/wagmi.ts
```diff
@@ -0,0 +1,43 @@
...
+export const config = createConfig({
+  chains: [activeChain],
+  transports: {
+    [mainnet.id]: http(mainnetRpcUrl),
+    [sepolia.id]: http(sepoliaRpcUrl),
+  },
+  connectors: [
+    injected(),
+    walletConnect({
+      projectId: WALLET_CONNECT_V2_PROJECT_ID,
```
volkyeth, line 29: shouldn't this one be prefixed with `VITE_`?

nekofar:
> when navigating directly to a Noun page, app breaks. e.g. http://localhost:3000/noun/1508 doesn't happen to all of them, so probably a specific kind of activity
> 
> ```
> Uncaught TypeError: Cannot convert a BigInt value to a number
>     at Array.sort (<anonymous>)
>     at useNounActivity (nounActivity.ts:294:6)
>     at ProfileActivityFeed (index.tsx:39:36)
>     ...
> ```

It should be ok now. https://github.com/nounsDAO/nouns-monorepo/pull/950/commits/69d728a5ed54e62a2b43ff1d4857d72a1ce38883

packages/nouns-webapp/src/wagmi.ts
nekofar: This is a constant. 
https://github.com/nounsDAO/nouns-monorepo/blob/c7c4d6f9039237665182cfc6d5264103bc2f94b6/packages/nouns-webapp/src/config.ts#L54-L55

volkyeth: lol my bad

we should probably get rid of those constants, tho. not a fan of needless indirections.
anyway, besides the point of this PR

packages/nouns-webapp/vite.config.ts
```diff
@@ -5,12 +5,11 @@ import { lingui } from '@lingui/vite-plugin';
...
 import svgr from 'vite-plugin-svgr';
 import Inspect from 'vite-plugin-inspect';
-import tailwindcss from '@tailwindcss/vite'
+import checker from 'vite-plugin-checker';
 
 // https://vite.dev/config/
 export default defineConfig({
   plugins: [
-    tailwindcss(),
     react({
```
volkyeth, line 13: was the tailwind plugin removed by accident?

nekofar: Nope, not fully integrated yet. Tailwind v4 needs a plugin, so I downgraded to v3 for now. v4 is still pretty new and has some compatibility quirks. Felt like it’s out of scope for this PR anyway; figured we can handle it separately if we ever ditch modular CSS and go full Tailwind.

nekofar:
@volkyeth I've added new environmental variables instead hardcoded addresses, you my need to set  `VITE_MAINNET_SUBGRAPH`.

packages/nouns-webapp/src/config.ts
```diff
@@ -70,15 +70,13 @@ const app: Record<SupportedChains, AppConfig> = {
...
-      'https://api.goldsky.com/api/public/project_cldf2o9pqagp43svvbk5u3kmo/subgraphs/nouns-sepolia-the-burn/0.1.0/gn',
+    subgraphApiUri: import.meta.env.VITE_SEPOLIA_SUBGRAPH ?? '',
     enableHistory: import.meta.env.VITE_ENABLE_HISTORY === 'true',
   },
   [mainnet.id]: {
     jsonRpcUri: createNetworkHttpUrl('mainnet'),
     wsRpcUri: createNetworkWsUrl('mainnet'),
-    subgraphApiUri:
-      'https://api.goldsky.com/api/public/project_cldf2o9pqagp43svvbk5u3kmo/subgraphs/nouns/prod/gn',
+    subgraphApiUri: import.meta.env.VITE_MAINNET_SUBGRAPH ?? '',
```
volkyeth, line 79: can you add those to .env.example(.local)

volkyeth:
@nekofar did you manage to get it running on Sepolia?
I tried it locally but doesn't seem to load at all (most likely bc the sepolia subgraph is dead)

Would be more confident merging this one if I could run more extensive tests on Sepolia, but I understand it might not be that easy rn

volkyeth:
CleanShot 2025-05-19 at 11 32 40
when a bid is placed, the loader never goes away (even after subgraph query refreshes and new bid shows up).

unsure if the problem was introduced by this PR or if it's a preexisting one, but should be a quick fix with `waitForTransactionReceipt`

volkyeth:
When trying to create a proposal or a candidate, right after adding an action to the prop, app breaks. relevant console errors:
```
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
[...]
hook.js:608 The above error occurred in the <CreateCandidatePage> component:

    at CreateCandidatePage (http://localhost:3000/src/pages/CreateCandidate/index.tsx:51:59)
    at StepProgress2 (http://localhost:3000/node_modules/.vite/deps/react-stepz.js?v=43b5aa5d:38:23)
    at RenderedRoute (http://localhost:3000/node_modules/.vite/deps/chunk-6L74X45C.js?v=43b5aa5d:5470:26)
```

packages/nouns-webapp/src/config.ts
nekofar: Added `VITE_MAINNET_SUBGRAPH` and `VITE_SEPOLY_SUBGRAPH` to `packages/nouns-webapp/.env.example.local`
Also updated the deployment scripts in `packages/nouns-subgraph/package.json` so you (or anyone) can spin up Mainnet or Sepolia subgraphs on The Graph for testing and dev
Just update your {user_id} in `VITE_MAINNET_SUBGRAPH` and `VITE_SEPOLIA_SUBGRAPH` after you deploy your subgraphs 🤘

nekofar:
> @nekofar did you manage to get it running on Sepolia? I tried it locally but doesn't seem to load at all (most likely bc the sepolia subgraph is dead)
> 
> Would be more confident merging this one if I could run more extensive tests on Sepolia, but I understand it might not be that easy rn
> 
> EDIT: managed to use this subgraph to run some local tests https://thegraph.com/explorer/subgraphs/AGjNFPHb2jWUyvSTnZqWn8LzPtLQLpjUXuZNVT812Hvr?view=Query&chain=arbitrum-one

I've deployed a new subgraph on Sepolia for testing. https://github.com/nounsDAO/nouns-monorepo/pull/950#discussion_r2096225972

nekofar:
> CleanShot 2025-05-19 at 11 32 40 when a bid is placed, the loader never goes away (even after subgraph query refreshes and new bid shows up).
> 
> unsure if the problem was introduced by this PR or if it's a preexisting one, but should be a quick fix with `waitForTransactionReceipt`

I think its related to redux states, I'll look into it.

nekofar:
> When trying to create a proposal or a candidate, right after adding an action to the prop, app breaks. relevant console errors:
> 
> ```
> Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
> [...]
> hook.js:608 The above error occurred in the <CreateCandidatePage> component:
> 
>     at CreateCandidatePage (http://localhost:3000/src/pages/CreateCandidate/index.tsx:51:59)
>     at StepProgress2 (http://localhost:3000/node_modules/.vite/deps/react-stepz.js?v=43b5aa5d:38:23)
>     at RenderedRoute (http://localhost:3000/node_modules/.vite/deps/chunk-6L74X45C.js?v=43b5aa5d:5470:26)
> ```

This one solved.

volkyeth:
@nekofar 
I think this is good enough to merge. We can extract the remaining non-critical issues to solve later.
wdyt?

I want to merge this one to unlock work on the rest of the issues without having to worry about nasty merge conflicts.
This should be the last of the big lifts, and following PRs should be more self-contained

packages/nouns-webapp/src/wagmi.ts
```diff
@@ -0,0 +1,42 @@
...
+const activeChainId = Number(CHAIN_ID);
+
+const activeChain =
+  pipe(
+    [mainnet, sepolia],
+    find(chain => chain.id === activeChainId),
+  ) ?? sepolia;
+
+const mainnetRpcUrl = process.env.VITE_MAINNET_JSONRPC ?? '';
+const sepoliaRpcUrl = process.env.VITE_SEPOLIA_JSONRPC ?? '';
```
volkyeth, line 17:
```suggestion
const mainnetRpcUrl = import.meta.env.VITE_MAINNET_JSONRPC ?? '';
const sepoliaRpcUrl = import.meta.env.VITE_SEPOLIA_JSONRPC ?? '';
```

those are currently not being populated

packages/nouns-webapp/src/wagmi.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const config = createConfig({
+  chains: [activeChain],
```
volkyeth, line 20: this is not working as intended.
When I'm connected to Sepolia on my wallet, wagmi is making contract calls to the Sepolia contract address on mainnet even when I have`VITE_CHAIN_ID=1` in my env, so the auction never loads

nekofar:
> > > When trying to create a proposal or a candidate, right after adding an action to the prop, app breaks. relevant console errors:
> > > [...]
> > > This one solved.
> 
> @nekofar actually, still happening when submitting a prop candidate. should be the exactly same problem you fixed for Proposals. take care of that one and I say we're good to go
> 
> created #1018 to reduce maintenance effort in the future

I've solved that as well  and I did some refactoring.

packages/nouns-webapp/src/wagmi.ts
nekofar: > this is not working as intended. When I'm connected to Sepolia on my wallet, wagmi is making contract calls to the Sepolia contract address on mainnet even when I have`VITE_CHAIN_ID=1` in my env, so the auction never loads. If I switch to mainnet it works as intended

Looks like there was a hard-coded chain id; I updated it. I’ve been testing auctions and proposals on Sepolia today and everything works (aside from the loading bid button, which I’ll sort). During migration I tried to mimic ethers behavior wherever possible so the UI and other site functionality wouldn’t break. We definitely need a refactor to simplify the code moving forward, but this branch has grown way too big for that 😅

volkyeth (approved):

PR #964, opened 2025-05-07
chore: improve testing and linting base

nekofar:
@volkyeth I've splitted some changes out of #950 so it be cleaner and also we have these changes reviewed and updated separately. 

* bumped webpack, globals, hardhat, netlify-cli, turbo, lint-staged, @typechain/ethers-v5, eslint-plugin-turbo, prettier-plugin-packagejson, mocha & types, chai & types, glob, vitest, typescript & TS-ESLint; removed ts-node, pinst, mocha, chai & their types
* updated general dependencies (nanoid range, multi-package upgrades)
* migrated ESLint to flat ES module config; added import-order, TS resolver, sonarjs, prettier, turbo plugins; tweaked rules, parserOptions, ignores; removed unused props
* migrated lint-staged to ES module config; updated tasks, excludes (pnpm-lock.yaml), fixed spacing
* organized imports in contracts, SDK, subgraph; reordered assets imports; replaced let with const in subgraph & tests
* swapped Mocha/Chai for Vitest in SDK & assets; added subgraph pretest script; marked asset tests as todo
* updated tsconfig include/exclude/tests, fixed formatting; added type field to package.json


packages/nouns-assets/test/util.test.ts
```diff
@@ -35,7 +31,7 @@ describe('@noun/assets utils', () => {
     });
   });
 
-  describe('getPseudorandomPart', () => {
+  describe.todo('getPseudorandomPart', () => {
```
volkyeth, line 34: are those test broken?
why the todo?

volkyeth (approved):

packages/nouns-assets/test/util.test.ts
volkyeth: ok, just checked and they were

PR #995, opened 2025-05-19
fix dead link NounsI18nProvider.tsx

dizer-ti:
Hey team—noticed a dead link, replaced it with a working URL

https://github.com/Uniswap/interface/blob/main/src/lib/i18n.tsx - old
https://github.com/Uniswap/interface/blob/main/packages/uniswap/src/i18n/i18n-setup.tsx - new link


volkyeth:
@dizer-ti that's probably the wrong reference, though, as this code does not have any resemblance to the one on the new link. we're better off removing that comment instead.

(for future reference, it's best to use [permalinks](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files) instead of mutable references)

volkyeth (approved):

PR #1029, opened 2025-05-20
ci: improve node pipeline workflow

nekofar:
Overhauls node ci workflow, merges all tests into one job, adds linting on PRs and runs on node LTS 22 and 24 so we don’t break prod later 😅


volkyeth (approved):
LGTM

Issue #1021, opened 2025-05-20
Link to sepolia etherscan on testnet app treasury link


volkyeth:
![Image](https://github.com/user-attachments/assets/e272ecf7-53d5-47c5-9b11-4ef7731ad8a1)

currently takes you to mainnet etherscan

Issue #1028, opened 2025-05-20
Auction won't load if wallet isn't connected to mainnet


volkyeth:
If the browser wallet is connected to any chain other than Ethereum mainnet, the auction won't load

![Image](https://github.com/user-attachments/assets/eeaa69a3-bbb5-4b21-bcf4-c8e6f9d4bad2)

PR #1031, opened 2025-05-20
fix: solve auction issue  by adding default chain

nekofar:
It's related to #1028. I've replaced `useChainId` by `defaultChain.id`, based in Wagmi configs. Loading auction should not be related to wallet, make sure clear application data, disconnect wallet, and provide error logs if you still have any issues.
Nouns DAO


nekofar:
Fixed lint errors in all the files I touched so CI finally chills; only sane way to keep it green over time 😅

nekofar:
I fixed #1021 while I was working on this, as it was related.

volkyeth:
Auction still won't load with:
- `VITE_CHAIN_ID=1`
- wallet connected to Sepolia (using Rabby wallet)

relevant logs:
(looks like bigint serialization issue)
```
index.tsx:121 A non-serializable value was detected in an action, in the path: `payload.timestamp`. Value: 1747768787n 
Take a look at the logic that dispatched this action:  {type: 'auction/appendBid', payload: {…}} 
(See https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants) 
(To allow non-serializable values see: https://redux-toolkit.js.org/usage/usage-guide#working-with-non-serializable-data)
```

```js
//prev state
{
    "account": {
        "activeAccount": "0x6a024f521f83906671e1a23a8B6c560be7e980F4"
    },
    "application": {
        "stateBackgroundColor": "#d5d7e1",
        "isCoolBackground": true,
        "alertModal": {
            "show": false
        }
    },
    "auction": {
        "bids": []
    },
    "logs": {},
    "pastAuctions": {
        "pastAuctions": []
    },
    "onDisplayAuction": {}
}
```
```js
// action
{
    "type": "auction/appendBid",
    "payload": {
        "nounId": "1521",
        "sender": "0xA7860E99e3ce0752D1ac53b974E309fFf80277C6",
        "value": "1500000000000000000",
        "extended": false,
        "transactionHash": "0x4f9bb57a36f288a86216b00d8aa6bfd529770b18c4fa8fdaf0ede83d1633a520",
        "transactionIndex": 293,
        "timestamp": "1747768787"
    }
}
```

```js
// next state
{
    "account": {
        "activeAccount": "0x6a024f521f83906671e1a23a8B6c560be7e980F4"
    },
    "application": {
        "stateBackgroundColor": "#d5d7e1",
        "isCoolBackground": true,
        "alertModal": {
            "show": false
        }
    },
    "auction": {
        "bids": []
    },
    "logs": {},
    "pastAuctions": {
        "pastAuctions": []
    },
    "onDisplayAuction": {}
}
```

volkyeth:
> Auction still won't load with

Update:
fixed problem of network switch modal not being displayed on d0837ff

auction still won't load, and I've identified the Wagmi cli generated calls are targeting the contract address from the currently connected network:
https://github.com/nekofar/nouns-monorepo/blob/d0837ffc33156f5f3458c7eafc82bcd1936372f5/packages/nouns-webapp/src/index.tsx#L84
this hook is failing bc it's calling the sepolia address on mainnet (using `VITE_CHAIN_ID=1`, wallet connected to Sepolia)
I'm investigating possible fixes

nekofar:
Let’s clear this up 🤔 do we really want users flipping their wallet to Sepolia on prod? I doubt it. So don’t trust the wallet’s network. Spin up a dedicated Sepolia endpoint (e.g. sepolia.nouns.wtf or testnet.nouns.wtf) for test, and keep the main site locked to mainnet.

nekofar:
Think auction glitch you mentioned boils down to redux. Site uses state not onchain data so switching network still pulls state from the previous chain and that’s what’s messing with the UI. I’ll dig into it later if it’s still acting up 🤔

volkyeth:
@nekofar
> Let’s clear this up 🤔 do we really want users flipping their wallet to Sepolia on prod? I doubt it. So don’t trust the wallet’s network. Spin up a dedicated Sepolia endpoint (e.g. sepolia.nouns.wtf or testnet.nouns.wtf) for test, and keep the main site locked to mainnet.

Agreed that the app network should be locked per deployment.
It should be possible to do by setting [`syncConnectedChain`](https://wagmi.sh/react/api/createConfig#syncconnectedchain) to false.
reported https://github.com/wevm/wagmi/issues/4680 since that should also work for wagmi-cli hooks

> Let’s pop up a modal or something to force users onto the right network. We’ll also able to lock the chain on every tx so, even if their wallet’s on the wrong chain, the tx still lands where it should. But works only on new wallets.

Yes, this is the current behaviour after d0837ff. It's good enough for now.

People will open this website all the time just to check the status of the auction, or find some link. It's pretty annoying for users using wallets that have a global chain setting instead of per-app to switch network all the time.

Ideally the user would be prompted to switch networks only when he initiated an onchain action.
But this improvement can be extracted to another issue.

> Think auction glitch you mentioned boils down to redux. Site uses state not onchain data so switching network still pulls state from the previous chain and that’s what’s messing with the UI. I’ll dig into it later if it’s still acting up 🤔

it's caused by https://github.com/wevm/wagmi/issues/4680
connected network has precedence over config network

packages/nouns-webapp/src/components/CandidateSponsors/SignatureForm.tsx
```diff
@@ -75,7 +76,8 @@ const SignatureForm = (props: Readonly<SignatureFormProps>) => {
   const [isTxSuccessful, setIsTxSuccessful] = useState(false);
   const [errorMessage, setErrorMessage] = useState<ReactNode>('');
 
-  const chainId = useChainId();
+  const chainId = defaultChain.id;
```
volkyeth, line 79: could use `CHAIN_ID` for consistency, and also it's more explicitly a static (build time) config. `defaultChain.id` has adds more indirections

nekofar: `defaultChain` isn’t just a chain id; it’s packed with metadata we can plug in later to DRY up our code, from explorer links to any other chain-specific details.

packages/nouns-webapp/src/index.tsx
```diff
@@ -60,28 +61,24 @@ const createRootReducer = () =>
...
-      // ... other middlewares ...
-    ),
-  );
-}
+const loggerMiddleware = createLogger();
 
-const store = configureStore({});
+export const store = configureStore({
+  reducer: createRootReducer(),
+  middleware: getDefaultMiddleware => getDefaultMiddleware().concat(loggerMiddleware),
```
volkyeth, line 68: can we please conditionally enable it with a ENV?
this makes the console too noisy...its harder to debug other stuff

packages/nouns-webapp/src/components/CandidateSponsors/SignatureForm.tsx
volkyeth: nitpick, but should probably pick one or another for consistency.
(CHAIN_ID is used right below this one)

packages/nouns-webapp/src/utils/etherscan.ts
```diff
@@ -1,16 +1,17 @@
...
-    case sepolia.id:
-      return 'https://sepolia.etherscan.io/';
-    default:
-      return 'https://etherscan.io/';
+  if (network === sepolia.id) {
+    return 'https://sepolia.etherscan.io/';
+  } else {
+    return 'https://etherscan.io/';
   }
 };
```
volkyeth, line 12: could use `defaultChain.blockExplorers.default.url`

packages/nouns-webapp/src/utils/etherscan.ts
```diff
@@ -33,15 +34,14 @@ export const buildEtherscanHoldingsLink = (address: string): string => {
...
 
 const getApiBaseURL = (network: number) => {
-  switch (network) {
-    case sepolia.id:
-      return 'https://api-sepolia.etherscan.io/';
-    default:
-      return 'https://api.etherscan.io/';
+  if (network === sepolia.id) {
+    return 'https://api-sepolia.etherscan.io/';
```
volkyeth, line 38: could use `defaultChain.blockExplorers.default.apiUrl`

packages/nouns-webapp/src/index.tsx
nekofar: It's done.

packages/nouns-webapp/src/utils/etherscan.ts
nekofar: It's done.

packages/nouns-webapp/src/utils/etherscan.ts
nekofar: It’s done

packages/nouns-webapp/src/components/CandidateSponsors/SignatureForm.tsx
nekofar: Fix this specific one, but if I update all of it's usage its going to trigger a-lot of lint errors, trying to keep PRs small as possible moving forward. 😹

PR #1038, opened 2025-05-22
refactor(webapp): remove Noun activity section

### LINKED ISSUE #1001: Remove Noun activity section
Image

This section doesn't really serve a purpose since most Nouners hold multiple tokens these days, so the history of a specific Noun is not very useful info.

removing this could significantly reduce the complexity of the main page, since there's a variant for the active auction and another for previous Nouns. They could be unified without this section.

volkyeth:
Closes #1001


volkyeth:
@nekofar can I count on you for reviews? It's lonely in here 🥲

PR #1047, opened 2025-05-23
Fixes auction navigation issue

### LINKED ISSUE #1039: Auction fails to transition from current auction to previous Noun on first try
When user is on the current auction/root `/` and clicks the left arrow to navigate to the previous Noun, the page quickly flashes to `/noun/{previousNounId}` and then goes back to `/noun/{currentNounId}`

It's most likely related to the redux logic on auctions
this might be the time to replace all of that with plain tanstack/query to simplify

volkyeth:
Fixes #1039 

- sets up graphql-codegen for use with tanstack/query
- sets up type-safe ide graphql query completions
- "forbids" usage of apollo/client (start migration to tanstack/query)
- fixes unserializable redux payload error

As you can see from the branch name, I started out intending to do a bigger refactor switching to plain `@tanstack/query` state management for auctions, but that's a huge can of worms.


nekofar:
Is it done? Since you set  ready, but I see some new commits  aswell.

nekofar:
Yeah, looks good. Might be a good idea to start throwing in some e2e tests at some point too.

